PR #245, opened 2020-09-24
Allow private settings files

ant0nsc:
Add the capability to not check in the complete `settings.yml` file, and fill in the missing ones via the file `InnerEyePrivateSettings.yml` file in the repository root.

### DIFF
--- a/InnerEye/Azure/secrets_handling.py
+++ b/InnerEye/Azure/secrets_handling.py
@@ -17,22 +17,14 @@ from InnerEye.Common import fixed_paths
 # is case sensitive.
 
 # The application key to access the subscription via ServicePrincipal authentication.
+from InnerEye.Common.fixed_paths import PRIVATE_SETTINGS_FILE
+
 APPLICATION_KEY = "APPLICATION_KEY"
 # The access key for the Azure storage account that holds the datasets.
 DATASETS_ACCOUNT_KEY = "DATASETS_ACCOUNT_KEY"
 
 # A list of all secrets that are stored in environment variables or local secrets files.
 SECRETS_IN_ENVIRONMENT = [APPLICATION_KEY, DATASETS_ACCOUNT_KEY]
-
-
-def set_environment_variables(variables: Dict[str, str]) -> None:
-    """
-    Creates an environment variable for each entry in the given dictionary. The dictionary key is the variable
-    name, it will be converted to uppercase before setting.
-    :param variables: The variable names and their associated values that should be set.
-    """
-    for name, value in variables.items():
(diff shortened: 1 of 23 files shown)


InnerEye/Scripts/submit_for_inference.py
```diff
@@ -162,12 +162,14 @@ def submit_for_inference(args: SubmitForInferenceConfig) -> Optional[Path]:
...
 
-def main(args: Optional[List[str]] = None) -> None:
+def main(args: Optional[List[str]] = None, project_root: Optional[Path] = None) -> None:
     """
     Main function.
     """
     logging_to_stdout()
-    submit_for_inference(SubmitForInferenceConfig.parse_args(args))
+    inference_config = SubmitForInferenceConfig.parse_args(args)
+    azure_config = AzureConfig.from_yaml(inference_config.settings, project_root=project_root)
```
Shruthi42, line 171: For consistency in the command line interface, could we allow users to override the AzureConfig parameters via the command line, the way we do in runner.py?

The same goes for report_structure_extremes.py and tensorboard_monitor.py.

InnerEye/Azure/secrets_handling.py
```diff
@@ -102,10 +94,53 @@ def throw_or_return_none(message: str) -> Optional[str]:
...
+        for key, value in dict2.items():
+            result[key] = value
+    return result
+
+
+def read_settings_yaml_file(yaml_file: Path) -> Dict[str, Any]:
     """
     Reads a YAML file, that is expected to contain an entry 'variables'. Returns the dictionary for the 'variables'
-    section of the file. If the file name is not given, an empty dictionary will be returned.
+    section of the file.
```
Shruthi42, line 143: Since we've changed the docstring, can we raise a ValueError if it is None?

Shruthi42 (dismissed):

InnerEye/Scripts/submit_for_inference.py
ant0nsc: If we add the full azure_config, it means that --help will no longer be really helpful, because that will bloat with 20 parameters that are not really relevant. I've added a --cluster argument though, as discussed over IM.

InnerEye/Azure/secrets_handling.py
ant0nsc: I'd rather not. I'd leave it to mypy to find cases where we call this function with an argument that does not comply with the type annotation.

jaysnanavati (approved):

Shruthi42 (approved):

PR #250, opened 2020-09-28
Improve test coverage for training recovery

ant0nsc:
Add a test for job recovery in the PR build, and individual unit tests


ant0nsc:
out of date, we have recovery test in AzureML now.

PR #258, opened 2020-09-30
Adding training metrics to metrics_aggregates for ScalarModels

melanibe:
Also adding training metrics to metrics_aggregates for ScalarModels


InnerEye/ML/visualizers/plot_cross_validation.py
```diff
@@ -182,6 +177,13 @@ def get_short_name(self, run_or_id: Union[Run, str]) -> str:
                 self.short_names[run_or_id] = run_id
         return self.short_names[run_id]
 
+    @property
+    def execution_mode_to_download(self) -> List[ModelExecutionMode]:
```
jaysnanavati, line 181: This could be a method that you can override separately in scalar config and segmentation config?

jaysnanavati (approved):

InnerEye/ML/visualizers/plot_cross_validation.py
```diff
@@ -182,6 +177,13 @@ def get_short_name(self, run_or_id: Union[Run, str]) -> str:
                 self.short_names[run_or_id] = run_id
         return self.short_names[run_id]
 
+    @property
```
ant0nsc, line 180: nitpicking: should be plural, execution_modes
nitpicking: I would not use a property here. Properties are, in my view, meant to read out simple data from an object.

InnerEye/ML/visualizers/plot_cross_validation.py
ant0nsc: Not sure that's worth it. And maybe we do actually want to download also training metrics for segmentation, if they are available.

ant0nsc (approved):

InnerEye/ML/visualizers/plot_cross_validation.py
melanibe: Metrics are always empty for training in segmentation it seems. I guess the reason is that it would be heavy to compute all of them?

InnerEye/ML/visualizers/plot_cross_validation.py
melanibe: updated

ant0nsc (approved):

InnerEye/ML/visualizers/plot_cross_validation.py
jaysnanavati: Overriding is not worth it? we are essentially doing the similar logic with the if right? I guess if we have a class hierarchy like this it might make sense to use it for these aspects perhaps?

melanibe: I don't know for me this attribute should be an attribute of `PlotCrossValConfig` not of the model itself as it merely concerns the behavior within the cross-validation processing

jaysnanavati (approved):

PR #275, opened 2020-10-08
Add ability to cache datasets using MonoAI CacheDataset + Refactoring

jaysnanavati:
- Add ability to cache datasets using MonoAI CacheDataset
- Remove custom implementations of Compose, to prefer using MonoAI transforms
- Refactor Segmentation, Scalar and Sequence datasets so that they can be cachable
- Migrate pre-processing functions in all datasets to use Transforms instead.


ant0nsc:
Closing for now, not actively worked on.

PR #294, opened 2020-10-21
Add numpy and hdf5 support to segmentation models

javier-alvarez:
- Add support for npy and npz arrays for segmentation models
- Add support for hdf5 files in segmentation models
- Change subjectIDs to strings


philippwitte:
To use hdf5 with the `SegmentationModelBase`, there are a few checks in `full_image_dataset.py` that I had to disable for now:

- Line 263 throws an error because it finds multiple file extensions. This is because my image path has the form `path_to_file.h5:field_name:channel:label`.

- Line 244 expects nifti format

- Line 350 is checking whether the hdf5 file exists at the image path, which throws an error due to the above naming convention

- Line 285 expects nifti format

docs/creating_dataset.md
```diff
@@ -19,7 +19,11 @@ InnerEye expects segmentation datasets to have the following structure:
  at minimum:
     * `subject`: A unique positive integer assigned to every patient
     * `channel`: The imaging channel or ground truth structure described by this row. 
-    * `filePath`: Path to the Nifti file for this scan or structure
+    * `filePath`: Path to the file for this scan or structure. We support nifti (nii.gz), numpy (npz) and hdf5(h5).
+        * For HDF5 path suffix with | separator
+            * For images <path>|<dataset_name>|<channel index>
+            * For segmentation binary <path>|<dataset_name>|<channel index>
+            * For segmentation multimap <path>|<dataset_name>|<channel index>|<multimap value>
```
ant0nsc, line 26: We also need to add here what assumption we make about the ordering of the array. Which index in the HDF5 array corresponds to X, Y, Z? Are we dropping any dimension? 
Segmentation binary and multimap should be better described.

docs/creating_dataset.md
```diff
@@ -19,7 +19,11 @@ InnerEye expects segmentation datasets to have the following structure:
  at minimum:
     * `subject`: A unique positive integer assigned to every patient
     * `channel`: The imaging channel or ground truth structure described by this row. 
-    * `filePath`: Path to the Nifti file for this scan or structure
+    * `filePath`: Path to the file for this scan or structure. We support nifti (nii.gz), numpy (npz) and hdf5(h5).
+        * For HDF5 path suffix with | separator
+            * For images <path>|<dataset_name>|<channel index>
+            * For segmentation binary <path>|<dataset_name>|<channel index>
```
ant0nsc, line 25: what happens if what we think is a binary segmentaiton is actually not?

ant0nsc (changes requested):
It's looking great overall, I've left comments where some polishing is still required.
I'd also suggest to improve test coverage for HDF5 loading. You are presently only testing the happy path (with the nifti-hdf5 conversion). If you make use of the mock libraries, you could also test the corner and failure cases (dataset_name does not exist, index out of range). Segmentation multimap loading is, I think, not tested either.

InnerEye/ML/utils/split_dataset.py
```diff
@@ -215,16 +214,16 @@ def from_proportions(df: pd.DataFrame,
             proportion_test=proportion_test
         )
         return DatasetSplits.from_subject_ids(df,
-                                              list(ranges[ModelExecutionMode.TRAIN]),
-                                              list(ranges[ModelExecutionMode.TEST]),
-                                              list(ranges[ModelExecutionMode.VAL]),
+                                              list(map(str, ranges[ModelExecutionMode.TRAIN])),
```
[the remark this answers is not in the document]
javier-alvarez: correct

InnerEye/ML/reports/notebook_report.py
```diff
@@ -48,7 +53,7 @@ def generate_notebook(template_notebook: Path, notebook_params: Dict, result_not
             resources=resources,
             notebook_name=str(result_notebook.stem)
         )
-    return result_notebook.with_suffix(resources['output_extension'])
+    return resources
```
ant0nsc, line 56: does nto need to return resources at all. Rather, it should return result_notebook.with_suffix(resources['output_extension'])

docs/creating_dataset.md
javier-alvarez: returns all 0s

ant0nsc (approved):

Shruthi42 (approved):

PR #295, opened 2020-10-22
Add python notebook and html for classification model reports

Shruthi42:
- During inference, write a file metrics.csv with per subject results.
- Adds a python notebook with the report for classification models.


InnerEye/Common/metrics_dict.py
```diff
@@ -538,11 +547,6 @@ def get_pr_auc(self, hue: str = DEFAULT_HUE_KEY) -> float:
             return np.nan
         predictions = self.get_predictions(hue)
         labels = self.get_labels(hue)
-        if predictions.shape[1] == 1 and labels.shape[1] == 1 and len(np.unique(labels)) == 1:
```
melanibe, line 541: Why removing that?

InnerEye/ML/model_testing.py
```diff
@@ -437,8 +440,27 @@ def test_epoch(test_epoch: int, run_recovery: Optional[RunRecovery]) -> Optional
...
+                csv_file = epoch_folder / METRICS_FILE_NAME
+
+                logging.info(f"Writing {data_split.value} metrics to file {str(csv_file)}")
+
+                # if we are running on the validation set, these may have been written during train time
+                if not csv_file.exists():
+                    os.makedirs(str(epoch_folder), exist_ok=False)
+                    df_logger = DataframeLogger(csv_file)
+
+                    # cross validation split index not relevant during test time
```
melanibe, line 454: How does this work if we are running a cross-validation run? One file per child run?

InnerEye/ML/model_testing.py
```diff
@@ -437,8 +440,27 @@ def test_epoch(test_epoch: int, run_recovery: Optional[RunRecovery]) -> Optional
         else:
             results[epoch] = epoch_result
 
+            if isinstance(epoch_result, ScalarMetricsDict):
+                epoch_folder = config.outputs_folder / get_epoch_results_path(epoch, data_split, model_proc)
+                csv_file = epoch_folder / METRICS_FILE_NAME
+
+                logging.info(f"Writing {data_split.value} metrics to file {str(csv_file)}")
+
+                # if we are running on the validation set, these may have been written during train time
```
melanibe, line 449: Add something like "else if running on the test set, create the metrics csv" ?

InnerEye/Common/metrics_dict.py
Shruthi42: This is already the default behavior of  sklearn's precision_recall_curve() - there's a test case that makes sure of this as well.

InnerEye/ML/model_testing.py
Shruthi42: When we have a set of cross validation runs, yes, this will create a csv per child run - this code doesn't support the cross validation case yet. On the other hand, if we run inference with checkpoints from a cross validation run, at that point we will have a single run and a single csv.

InnerEye/ML/model_testing.py
Shruthi42: Changed the comment to clarify.

InnerEye/Common/metrics_dict.py
```diff
@@ -490,6 +490,15 @@ def get_accuracy_at05(self, hue: str = DEFAULT_HUE_KEY) -> float:
         return binary_classification_accuracy(model_output=self.get_predictions(hue=hue),
                                               label=self.get_labels(hue=hue))
 
+    @classmethod
+    def get_optimal_idx(cls, fpr: np.ndarray, tpr: np.ndarray) -> np.ndarray:
+        """
+        Given a list of FPR and TPR values corresponding to different thresholds, compute the index which corrsponds
```
javier-alvarez, line 496: typo?

javier-alvarez (commented):
Looks good

melanibe (dismissed):
👍

InnerEye/Common/metrics_dict.py
Shruthi42: Fixed

melanibe (approved):

javier-alvarez (approved):

PR #325, opened 2020-11-24
Print image type (TP, FP, TN, FN) for every image in classification report

Shruthi42:


InnerEye/ML/reports/classification_report.py
```diff
@@ -321,7 +321,8 @@ def plot_image_for_subject(subject_id: str,
                            dataset_file_column: str,
                            dataset_dir: Path,
                            im_size: Tuple,
-                           model_output: float) -> None:
+                           model_output: float,
+                           header: Optional[str]) -> None:
```
ant0nsc, line 325: docu for `header` missing

InnerEye/ML/reports/classification_report.py
```diff
@@ -333,6 +334,8 @@ def plot_image_for_subject(subject_id: str,
     :param model_output: The predicted value for this image
     """
     print_header("", level=4)
+    if header:
+        print_header(header, level=4)
```
ant0nsc, line 338: Wouldn't it make more sense to have `print_header(header or "", level=4)`? At the moment, you'd get 2 prints?

ant0nsc (dismissed):

InnerEye/ML/reports/classification_report.py
Shruthi42: The empty string is just spacing to format it nicely.

javier-alvarez (approved):

PR #337, opened 2020-12-08
Checkpoint handling in Pytorch Lightning

Shruthi42:
Makes changes to how we save, load and find checkpoints.

Removes params
 * test_start_epoch
 * test_diff_epochs
 * test_step_epochs
 * save_start_epoch
 
Replaces `register_model_only_for_epoch` with `only_register_model`

In PR build, jobs TrainInAzureML and TrainEnsemble need the full env to be set up (including pytorch) to run tests.


InnerEye/ML/model_training.py
```diff
@@ -53,11 +53,20 @@ def model_train(config: ModelConfigBase,
...
-        filename='best_val_loss_checkpoint',
-        monitor=f"{VALIDATION_PREFIX}{MetricType.LOSS.value}",
-        save_last=True)
+    # Get the path to the checkpoint to recover from
+    checkpoint_path = checkpoint_handler.get_recovery_path_train()
+
+    last_checkpoint_callback = ModelCheckpoint(dirpath=str(config.checkpoint_folder),
+                                               filename='best_val_loss_checkpoint',
+                                               monitor=f"{VALIDATION_PREFIX}{MetricType.LOSS.value}",
+                                               save_last=True)
```
ant0nsc, line 62: Maybe we should break it down even further: One checkpoint handler that only saves the last epoch's checkpoint, and one that monitors validation loss.

InnerEye/ML/model_training.py
```diff
@@ -53,11 +53,20 @@ def model_train(config: ModelConfigBase,
...
+
+    recovery_checkpoint_callback = ModelCheckpoint(dirpath=str(config.checkpoint_folder),
```
ant0nsc, line 64: I think we are saving ALL checkpoints here, right? I don't think we want that, this consume a huge amount of memory.

ant0nsc (approved):

PR #348, opened 2021-01-04
Check more locations for dataset file and fail if comparison files mi…

JonathanTripp:
Searching for "dataset.csv" and "metrics.csv" used ValueError in the except block, but Azure ML now throws UserErrorException 
if the file does not exist, so add that to the list to check.

Also, if the files are missing and they are expected then raise a new Exception and do not silently fail in the MLRunner.

- [ ] Add unit tests for all functions that you introduced or modified.
- [ ] Run PyCharm's code cleanup tools on your Python files.
- [ ] Link the correct GitHub issue for tracking.
- [ ] Update the [Changelog](CHANGELOG.md) file: Describe your change in terms of 
- [ ] When merging your PR, replace the default merge message with a description of your PR,


InnerEye/ML/baselines_util.py
```diff
@@ -203,6 +205,7 @@ def get_comparison_baselines(outputs_folder: Path, azure_config: AzureConfig,
                 # noinspection PyUnresolvedReferences
                 if path is not None and not path.exists():
                     logging.warning("    ... but it does not exist")
```
ant0nsc, line 207: this whole block from line 202 onwards, not sure this adds much value.  Do we not yet have all the information in the logs, from line 191 and 179? Also, in which cases is `download__outputs_from_run` going to return something that is not None, but the file does not actually exist? This would simplify the checks.

Tests/AfterTraining/test_after_training.py
```diff
@@ -90,6 +91,26 @@ def test_model_file_structure(test_output_dirs: OutputFolderForTests) -> None:
         pytest.fail(f"{len(missing)} files in the registered model are missing: {missing[:5]}")
 
 
+@pytest.mark.after_training
+def test_get_comparison_data(test_output_dirs: OutputFolderForTests) -> None:
```
ant0nsc, line 95: this test is a good start, but prone to run out of sync with the actual code. Can you refactor `get_comparison_baselines` such that you can test the downloading of files in isolation?

ant0nsc (changes requested):

ant0nsc (approved):

Shruthi42 (approved):

PR #363, opened 2021-01-18
Implement grouped dataset splits and cross-validation

dccastro:
This PR adds the ability to specify a `group_column` in addition to the primary `subject_column` when creating `DatasetSplits`. If given, this ensures that subjects within each group cannot be in separate training/test/validation sets or cross-validation folds.

### DIFF
--- a/InnerEye/ML/utils/split_dataset.py
+++ b/InnerEye/ML/utils/split_dataset.py
@@ -30,9 +30,18 @@ class DatasetSplits:
 
     def __post_init__(self) -> None:
         common_util.check_properties_are_not_none(self)
+
+        def pairwise_intersection(*collections):
+            """Returns any element that appears in more than one collection."""
+            from itertools import combinations
+            intersection = set()
+            for col1, col2 in combinations(map(set, collections), 2):
+                intersection |= col1 & col2
+            return intersection
+
         # perform dataset split validity assertions
         unique_train, unique_test, unique_val = self.unique_subjects()
-        intersection = set.intersection(set(unique_train), set(unique_test), set(unique_val))
+        intersection = pairwise_intersection(unique_train, unique_test, unique_val)
 
         if len(intersection) != 0:
             raise ValueError("Train, Test, and Val splits must have no intersection, found: {}".format(intersection))
--- a/InnerEye/ML/utils/split_dataset.py
+++ b/InnerEye/ML/utils/split_dataset.py
@@ -272,7 +272,8 @@ class DatasetSplits:
                          train_ids: Sequence[str],
                          test_ids: Sequence[str],
                          val_ids: Sequence[str],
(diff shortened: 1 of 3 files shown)


Shruthi42 (commented):
For consistency across configs, we should consider having a documented default for the group column name in the csv. For example, we have a default subject column name for segmentation dataset csv files, and we have a parameter in the config which is used to specify the subject column name in scalar datasets.

InnerEye/ML/utils/split_dataset.py
```diff
@@ -26,17 +26,36 @@ class DatasetSplits:
...
     subject_column: str = CSV_SUBJECT_HEADER
+    group_column: Optional[str] = None
     allow_empty: bool = False
 
     def __post_init__(self) -> None:
         common_util.check_properties_are_not_none(self)
+
+        def pairwise_intersection(*collections: Iterable) -> Set:
+            """Returns any element that appears in more than one collection."""
+            from itertools import combinations
```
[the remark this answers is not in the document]
dccastro: Not really; well spotted! I'll move it to the top.

dccastro:
I thought about adding a default group column name, but couldn't immediately come up with a good catch-all solution... I believe in most cases that would be `None` (i.e. no grouping). In other cases, we might want `group_column` to refer to subject ID, while `subject_column` points to image/series ID instead. The other obvious use-case I could see is grouping by data source/institution/hospital/etc. Do you have any suggestions?

Shruthi42 (approved):

InnerEye/ML/utils/split_dataset.py
```diff
@@ -183,48 +206,87 @@ def get_subject_ranges_for_splits(population: Sequence[str],
...
 
+    @staticmethod
+    def _from_split_keys(df: pd.DataFrame,
+                         train_keys: Sequence[str],
+                         test_keys: Sequence[str],
+                         val_keys: Sequence[str],
+                         *,  # make column names keyword-only arguments to avoid mistakes when providing both
+                         key_column: str,
+                         subject_column: str,
+                         group_column: Optional[str]) -> DatasetSplits:
```
ant0nsc, line 217: both column names should have "" as the default value, and group_column=None

InnerEye/ML/utils/split_dataset.py
```diff
@@ -183,48 +206,87 @@ def get_subject_ranges_for_splits(population: Sequence[str],
...
+        """
+        Takes a slice of values from each data split train/test/val for the provided keys.
```
ant0nsc, line 219: What's a slice of values?

ant0nsc (approved):

InnerEye/ML/utils/split_dataset.py
dccastro: I just adapted this docstring from `from_subject_ids()`.

PR #368, opened 2021-01-20
Documentation updates

csiebler:
Fixed minor confusions in documentation about instance sizes for AML and also AML Dataset configuration in Models


docs/building_models.md
```diff
@@ -64,12 +64,12 @@ class HeadAndNeck(HeadAndNeckBase):
...
         super().__init__(
             ground_truth_ids=["parotid_l", "parotid_r", "smg_l", "smg_r", "spinal_cord"]
-            azure_dataset_id="id-of-your-blob-containing-prostate-data")
+            azure_dataset_id="name-of-your-AML-dataset-with-prostate-data")
 ```
 
 ### Training a new model
 
-* Set up your model configuration as above.
+* Set up your model configuration as above and update `azure_dataset_id` to the name of your Dataset in the AML workspace.
```
ant0nsc, line 72: Add: It is enough to put your dataset into blob storage. The dataset should be a contained in a folder at the root of the `datasets` container. The InnerEye runner will check if there is a dataset in the AzureML workspace already, and if not, generate it directly from blob storage.

docs/setting_up_aml.md
```diff
@@ -64,8 +64,10 @@ You can skip this if you have chosen automatic deployment above.
 
 We recommend using [low priority](https://docs.microsoft.com/en-us/azure/batch/batch-low-pri-vms) clusters, since 
 they only cost a fraction of the dedicated VMs.
-As a reference, the Prostate model and the Head and Neck model require VMs with 4 GPUs with at least 16GB of memory
-per GPU, for example `Standard_ND24s`, `Standard_NC24s_v3` or `Standard_NC24s_v2`.
+As a reference:
+* The Prostate model and the Head and Neck model require VMs with 4 GPUs with at least 16GB of memory
```
ant0nsc, line 68: VMs with 4GPUs with at least 24GB of memory (ND24s) are required for all models, in fact.
Maybe here also say what to do if not fitting? Reduce train_batch_size, and if that's not enough, reduce train_crop_size and test_crop_size. Adjust inference_stride_size accordingly

ant0nsc (changes requested):

docs/setting_up_aml.md
csiebler: Makes, sense, I will need to check, as lowering the train_batch_size (8 --> 4) did not fix it for me

mkoivi-ms: Maybe ND24s should be mentioned as an example, to get the reader started quickly without having to analyze compute cluster SKUs.

ant0nsc (approved):

javier-alvarez (approved):

ant0nsc:
All good.

PR #391, opened 2021-02-08
Support Multilabel classification models with png files

javier-alvarez:
- This PR adds support for multilabel classification models. Multiclass models (where labels are mutually exclusive) are not supported. Changes made:
  - `dataset.csv` can include multiple labels per sample
  - Loss functions changed to support multilabel classification tasks
  - Extra report for multilabel classification tasks
- Add support for png images


InnerEye/ML/dataset/scalar_dataset.py
```diff
@@ -28,38 +28,44 @@
...
+    :param num_classes: Number of expected classes
     :return:
     """
-    if isinstance(label_string, float):
-        if math.isnan(label_string):
-            # When loading a dataframe with dtype=str, missing values can be encoded as NaN, and get into here.
-            return label_string
-        else:
-            raise ValueError(f"Subject {sample_id}: Unexpected float input {label_string} - did you read the "
-                             f"dataframe column as a string?")
```
ant0nsc, line 50: this means the code is no longer backwards compatible. How are missing values handled now?

InnerEye/ML/dataset/scalar_dataset.py
```diff
@@ -28,38 +28,44 @@
...
+
+    if '|' in label_string or label_string.isdigit():
+        if num_classes == 1:
```
ant0nsc, line 46: 1 class does not make sense?

InnerEye/ML/dataset/scalar_dataset.py
```diff
@@ -28,38 +28,44 @@
...
 
-def extract_label_classification(label_string: Union[str, float], sample_id: str) -> Union[float, int]:
+def extract_label_classification(label_string: str, sample_id: str, num_classes: int) -> List[float]:
     """
     Converts a string from a dataset.csv file that contains a model's label to a scalar.
     The function maps ["1", "true", "yes"] to 1, ["0", "false", "no"] to 0.
     If the entry in the CSV file was missing (no string given at all), it returns math.nan.
     :param label_string: The value of the label as read from CSV via a DataFrame.
     :param sample_id: The sample ID where this label was read from. This is only used for creating error messages.
+    :param num_classes: Number of expected classes
```
ant0nsc, line 41: Can you add more docu? What is expected here for binary classification, what for regression?

InnerEye/ML/dataset/scalar_dataset.py
```diff
@@ -28,38 +28,44 @@
...
+            return [0.0]
+
         raise ValueError(f"Subject {sample_id}: Label string not recognized: '{label_string}'")
     else:
-        return math.nan
+        return [math.nan]
 
 
-def extract_label_regression(label_string: Union[str, float], sample_id: str) -> Union[float, int]:
+def extract_label_regression(label_string: Union[str, float], sample_id: str, num_classes: int) -> List[float]:
```
ant0nsc, line 68: why do we have num_classes as an argument here at all? it is not used apart from doing a check?

InnerEye/ML/dataset/scalar_dataset.py
```diff
@@ -150,10 +158,12 @@ def load_single_data_source(subject_rows: pd.DataFrame,
                             categorical_data_encoder: Optional[CategoricalToOneHotEncoder] = None,
                             metadata_columns: Optional[Set[str]] = None,
                             is_classification_dataset: bool = True,
+                            num_classes: int = 1,
                             sequence_position_numeric: Optional[int] = None) -> T:
     """
     Converts a set of dataset rows for a single subject to a ScalarDataSource instance, which contains the
     labels, the non-image features, and the paths to the image files.
+    :param num_classes: number of classes
```
ant0nsc, line 166: for all places where num_classes is expected, there should be a consistent documentation that says what is expected in binary classification, multi-class, multi-label, regression.

InnerEye/ML/configs/classification/DummyMulticlassClassification.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def __init__(self) -> None:
+        num_epochs = 4
+        super().__init__(
+            local_dataset=full_ml_test_data_path("classification_data_multiclass"),
+            image_channels=["blue"],
+            image_file_column="path",
+            label_channels=["blue"],
+            class_names=["class0", "class1", "class2", "class3", "class4"],
+            # labels_exclusive=False,
```
ant0nsc, line 25: is there test coverage for the two cases labels_exclusive=True and =False?

InnerEye/ML/scalar_config.py
```diff
@@ -102,6 +102,9 @@ def scale(labels: List) -> Union[List, Any]:
 
 
 class ScalarModelBase(ModelConfigBase):
+    class_names: List[str] = param.List(class_=str,
+                                        default=["Default"],
```
ant0nsc, line 106: "Default" should re-use the existing constant for default hue name

InnerEye/ML/scalar_config.py
```diff
@@ -102,6 +102,9 @@ def scale(labels: List) -> Union[List, Any]:
...
+                                        doc="Class names.")
```
ant0nsc, line 107: docu should say way more - in which cases is that needed, and in which not?

Tests/ML/reports/test_classification_report.py
```diff
@@ -42,8 +45,8 @@ def test_generate_classification_report(test_output_dirs: OutputFolderForTests)
 def test_get_results() -> None:
     reports_folder = Path(__file__).parent
     test_metrics_file = reports_folder / "test_metrics_classification.csv"
-
-    results = get_results(test_metrics_file)
+    hues = "Default"
```
ant0nsc, line 48: re-use constant

Tests/ML/reports/test_classification_report.py
```diff
@@ -58,15 +61,15 @@ def test_functions_with_invalid_csv(test_output_dirs: OutputFolderForTests) -> N
     # Duplicate a subject
     with open(invalid_metrics_file, "a") as file:
         file.write("Default,1,5,1.0,1,-1,Test")
-
+    hue = "Default"
```
ant0nsc, line 64: re-use constant

InnerEye/ML/reports/classification_report.ipynb
```diff
@@ -205,8 +221,11 @@
    "outputs": [],
    "source": [
     "if val_metrics_csv.is_file() and test_metrics_csv.is_file():\n",
-    "    print_k_best_and_worst_performing(val_metrics_csv=val_metrics_csv, test_metrics_csv=test_metrics_csv, \n",
-    "                                      k=number_best_and_worst_performing)"
+    "    for hue in config.class_names:\n",
```
ant0nsc, line 224: This means that even for binary classification we get a header "Class=Default", which looks out of place.

InnerEye/ML/reports/classification_report.ipynb
```diff
@@ -226,9 +245,12 @@
    "source": [
     "if val_metrics_csv.is_file() and test_metrics_csv.is_file() and dataset_csv_path.is_file() and \\\n",
     " dataset_subject_column and dataset_file_column:\n",
-    "    plot_k_best_and_worst_performing(val_metrics_csv=val_metrics_csv, test_metrics_csv=test_metrics_csv, \n",
+    "    for hue in config.class_names:\n",
```
ant0nsc, line 248: This means that even for binary classification we get a header "Class=Default", which looks out of place.

InnerEye/ML/reports/classification_report.py
```diff
@@ -49,15 +49,17 @@ class ReportedMetrics(Enum):
     FalseNegativeRate = "false_negative_rate"
 
 
-def get_results(csv: Path) -> LabelsAndPredictions:
+def get_results(csv: Path, hue: str) -> LabelsAndPredictions:
```
ant0nsc, line 52: can we retire this weird name "hue"? it will die in the rest of the codebase soon.

InnerEye/ML/configs/classification/DummyMulticlassClassification.py
ant0nsc: I see no other mention of that in the PR, so I presume it is dead code. This means that the code is always assuming that this is a multi-label classification task, and will allow multiple labels per sample. I think this should be more carefully controlled, we don't want anyone to accidentally feed in labels that are not mutually exclusive into a model that assumes exclusion.

Tests/ML/models/test_scalar_model.py
```diff
@@ -134,6 +135,57 @@ def test_train_classification_model(test_output_dirs: OutputFolderForTests) -> N
...
+        assert metric.value in model_training_result.val_results_per_epoch[0], f"{metric.value} not in validation"
+
+    actual_train_loss = model_training_result.get_metric(is_training=True, metric_type=MetricType.LOSS.value)
+    actual_val_loss = model_training_result.get_metric(is_training=False, metric_type=MetricType.LOSS.value)
+    actual_lr = model_training_result.get_metric(is_training=True, metric_type=MetricType.LEARNING_RATE.value)
+    assert actual_train_loss == pytest.approx(expected_train_loss, abs=1e-6), "Training loss"
+    assert actual_val_loss == pytest.approx(expected_val_loss, abs=1e-6), "Validation loss"
+    assert actual_lr == pytest.approx(expected_learning_rates, rel=1e-5), "Learning rates"
+    test_results = model_testing.model_test(config, ModelExecutionMode.TRAIN,
+                                            checkpoint_handler=checkpoint_handler)
```
ant0nsc, line 185: There should be assertions on the metrics that come out, too

ant0nsc (changes requested):
- There is no documentation in the MD files about how to use a multi-class model. Similarly, there should be a description for multi-label case.
- The metrics that we are presently computing make sense (I think) for multi-label, but not for multi-class with mutually exclusive labels.

InnerEye/ML/reports/classification_report.ipynb
javier-alvarez: Should we force users to give classes always a name?

InnerEye/ML/dataset/scalar_dataset.py
javier-alvarez: There some function that gets created expecting this interface :/

InnerEye/ML/dataset/scalar_dataset.py
javier-alvarez: I think this is independent of the type with this changes everything is multiclass there is not binary anymore

InnerEye/ML/configs/classification/DummyMulticlassClassification.py
javier-alvarez: yes this is a good point let me take a look

InnerEye/ML/reports/classification_report.ipynb
javier-alvarez: I think this is not a big problem as use can always rename the class_names

InnerEye/ML/reports/classification_report.py
javier-alvarez: Yes but maybe not as part of this PR

InnerEye/ML/utils/supervised_criterion.py
```diff
@@ -74,15 +75,13 @@ def get_positive_class_weights(self) -> torch.Tensor:
...
-        labels = list(self._class_counts.keys())
-        if sorted(labels) != [0.0, 1.0]:
-            if labels == [1.0] or labels == [0.0]:
-                return torch.tensor(1.0)
-            else:
-                raise ValueError(f"Expected one-hot encoded binary label."
-                                 f"Found labels {self._class_counts.keys()}")
-        else:
+        if self.num_classes == 1:
             return torch.tensor(float(self._class_counts[0.0]) / self._class_counts[1.0], dtype=torch.float32)
```
Shruthi42, line 79: Why is this (# of Class 0 samples) / (# of Class 1 samples) in the binary case and (# of Class 'i' samples) / (total # of samples) in the multilabel case? It should be consistent.

InnerEye/ML/scalar_config.py
```diff
@@ -41,6 +42,7 @@ class EnsembleAggregationType(Enum):
 @unique
 class ScalarLoss(Enum):
     BinaryCrossEntropyWithLogits = "BinaryCrossEntropyWithLogits"
+    # WeightedCrossEntropyWithLogits does not work with 1 class_names, use 2 class_names and 2 outputs to allow weights
```
Shruthi42, line 45: Is this still true? There is logic in BinaryCrossEntropyWithLogits to handle the single class case.

InnerEye/ML/scalar_config.py
```diff
@@ -41,6 +42,7 @@ class EnsembleAggregationType(Enum):
...
     WeightedCrossEntropyWithLogits = "WeightedCrossEntropyWithLogits"
     MeanSquaredError = "MeanSquaredError"
```
Shruthi42, line 47: For the case where labels are exclusive, we should have an option to use categorical cross entropy loss.

InnerEye/ML/scalar_config.py
javier-alvarez: This should be removed. At some point I considered not supporting this case

ant0nsc (changes requested):
This is in good shape, but as discussed, I think it would improve maintainability of you could make less use of raw dataframes at report generation.

InnerEye/ML/run_ml.py
```diff
@@ -719,21 +720,37 @@ def get_epoch_path(mode: ModelExecutionMode) -> Path:
...
+                    generate_classification_notebook(
+                        result_notebook=reports_dir / get_ipynb_report_name(config.model_category.value),
+                        config=config,
+                        train_metrics=path_to_best_epoch_train,
+                        val_metrics=path_to_best_epoch_val,
+                        test_metrics=path_to_best_epoch_test,
+                        dataset_csv_path=config.local_dataset / DATASET_CSV_FILE_NAME
+                        if config.local_dataset else None)
+
+                    if len(config.class_names) > 1:
```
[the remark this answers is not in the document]
Shruthi42: Yes, the standard report we create for the binary classification task, and another report which considers combinations of label classes and metrics specific to the multilabel case.

InnerEye/ML/dataset/scalar_dataset.py
```diff
@@ -747,13 +784,21 @@ def get_labels_for_imbalanced_sampler(self) -> List[float]:
...
+    def get_class_counts(self) -> Dict[float, int]:
         """
-        Return class weights that are proportional to the inverse frequency of label counts.
+        Return the label counts.
         :return: Dictionary of {"label": count}
         """
-        all_labels = [item.label.item() for item in self.items]  # [N, 1]
-        return dict(Counter(all_labels))
+        if len(self.args.class_names) == 1:
+            return dict(Counter([item.label.item() for item in self.items if item.label.item() != 0]))
```
ant0nsc, line 793: I don't get that part. If you are feeding only labels of class 1 into `Counter`, you would only count positive labels, right? 
Also, is there a test for that function?

ant0nsc: Would also be helpful to have more detailed docu for the multilabel case, in particularly clarifying if the dictionary has class indices or names as keys.

InnerEye/ML/dataset/scalar_dataset.py
```diff
@@ -747,13 +784,21 @@ def get_labels_for_imbalanced_sampler(self) -> List[float]:
         """
         return [item.label.item() for item in self.items]
 
-    def get_class_counts(self) -> Dict:
+    def get_class_counts(self) -> Dict[float, int]:
```
ant0nsc, line 787: is the type annotation correct? In the multi-label case, it looks like Dict[int,int] to me.

InnerEye/ML/dataset/scalar_sample.py
```diff
@@ -175,6 +164,47 @@ def is_valid(self) -> bool:
...
+        :param file: Image filepath relative to the dataset folder
+        :param root_path: The root path where all channel files for images are expected. This is ignored if
+        file_mapping is given.
+        :param file_mapping: A mapping from a file name stem (without extension) to its full path.
+        """
+        if file is None:
+            raise ValueError("When loading images, channel_files should no longer contain None entries.")
+        elif file_mapping:
+            if file in file_mapping:
+                return file_mapping[str(file)]
```
ant0nsc, line 200: we're doing str() on an argument that is already str?

InnerEye/ML/reports/classification_report.py
```diff
@@ -49,19 +52,32 @@ class ReportedMetrics(Enum):
...
-def get_results(csv: Path) -> LabelsAndPredictions:
+def read_csv_and_filter_prediction_target(csv: Path, prediction_target: str) -> pd.DataFrame:
     """
-    Given a CSV file, reads the subject IDs, ground truth labels and model outputs for each subject.
-    NOTE: This CSV file should have results from a single epoch, as in the metrics files written during inference, not
-    like the ones written while training.
+    Given one of the csv files written during inference time, read it and select only those rows which belong to the
+    given prediction_target. Also check that there is only a single entry per prediction_target per subject in the file.
+    The csv must have at least the following columns (defined in the LoggingColumns enum):
+    LoggingColumns.Hue, LoggingColumns.Patient, LoggingColumns.Label, LoggingColumns.ModelOutput.
```
ant0nsc, line 60: why is label and modeloutput mandatory, if the code does not use them?

InnerEye/ML/scalar_config.py
```diff
@@ -102,6 +103,17 @@ def scale(labels: List) -> Union[List, Any]:
...
+                                        default=[MetricsDict.DEFAULT_HUE_KEY],
+                                        bounds=(1, None),
+                                        doc="The label names for each label class in the dataset and model output "
+                                            "in the case of binary and multi-label classification tasks."
+                                            "The order of the names should match the order of label class indices "
+                                            "in dataset.csv"
+                                            "For multi-label classification, this field is required."
+                                            "For binary classification, this field must be a list of size 1, and "
+                                            "is by default ['Default'], but can optionally be set to a more descriptive "
+                                            "name for the positive class.")
```
ant0nsc, line 116: is it tested that the code works if len(class_names)==1, but it's not the default?

ant0nsc (dismissed):

InnerEye/ML/reports/classification_report.py
Shruthi42: Moved this comment to the appropriate function docstrings.

InnerEye/ML/dataset/scalar_dataset.py
Shruthi42: The intended behavior for binary classification tasks is to return the positive class count. I've added more details in the docstring. There are tests for this function in `test_scalar_dataset.py`.

InnerEye/ML/reports/classification_multilabel_report.py
```diff
@@ -0,0 +1,173 @@
...
+
+    return label_set
+
+
+def get_dataframe_with_exact_label_matches(metrics_df: pd.DataFrame,
+                                           prediction_target_set_to_match: List[str],
+                                           all_prediction_targets: List[str],
+                                           thresholds_per_prediction_target: List[float]) -> pd.DataFrame:
+    """
+    Given a set of prediction targets, for each sample find
```
melanibe, line 54: Am not sure i understand what "a set of prediction targets" is here

InnerEye/ML/reports/classification_multilabel_report.py
```diff
@@ -0,0 +1,173 @@
...
+    df[LoggingColumns.Hue.value] = ["|".join(prediction_target_set_to_match)] * len(df)
+    return df
+
+
+def get_labels_and_predictions_for_prediction_target_set(csv: Path,
+                                                         prediction_target_set_to_match: List[str],
+                                                         all_prediction_targets: List[str],
+                                                         thresholds_per_prediction_target: List[float]) -> LabelsAndPredictions:
+    """
+    Given a CSV file, generate a set of labels and model predictions for the combination of prediction targets.
```
melanibe, line 118: same here

InnerEye/ML/reports/classification_multilabel_report.py
```diff
@@ -0,0 +1,173 @@
...
+    Given a set of prediction targets, for each sample find
+        (i) if the set of ground truth labels matches this set exactly,
+        (ii) if the predicted model outputs (after thresholding) match this set exactly
+
+    Generates an output dataframe with the rows:
+    LoggingColumns.Patient, LoggingColumns.Label, LoggingColumns.ModelOutput, LoggingColumns.Hue
+
+    The output dataframe is generated according to the following rules:
+      - LoggingColumns.Patient: For each sample, the sample id is copied over into this field
+      - LoggingColumns.Label: For each sample, this field is set to 1 if the set of ground truth labels for the sample
```
melanibe, line 63: Am really confused was it the difference between set of ground truth labels and set of prediction targets?

InnerEye/ML/utils/supervised_criterion.py
```diff
@@ -56,10 +56,28 @@ def forward_minibatch(self, output: Any, target: Any, **kwargs: Any) -> Any:
...
 
-    def __init__(self, class_counts: Optional[Dict[float, float]] = None, **kwargs: Any):
+    def __init__(self, num_classes: int,
+                 class_counts: Optional[Dict[float, int]] = None,
+                 num_train_samples: Optional[int] = None,
+                 **kwargs: Any):
+        """
+        :param num_classes: The number of classes the model predicts. For binary classification num_classes is one
+                            and for multi-label classification tasks num_classes will be greater than one.
+                            Note that in multi-label tasks, the negative class (all other ground truth classes being
```
melanibe, line 66: not sure i get what "the negative class should not be a separate class means"

Shruthi42: I've moved this bit to `creating_dataset.md`, it makes more sense there.

docs/creating_dataset.md
```diff
@@ -270,4 +274,52 @@ In this example, `weight` is a scalar feature read from the csv, and `class` is
 **Filtering on channels**: This example also shows why filtering values by channel is useful: In this example, each subject has 2 images taken at 
  different times with different label values. By using `label_channels=["image_time_2"]`, we can use the label associated with
  the second image for all subjects.
- 
\ No newline at end of file
+
+#### Multi-label classification datasets
+Classification datasets can be multi-label, i.e. they can have more than one label associated with every sample.
+In this case, in the label column, separate the (numerical) ground truth labels with a pipe character (`|`) to 
+provide multiple ground truth labels for the sample.
```
melanibe, line 281: Should we add here somekind of warning: at this point we support multi-label classification only (i.e. one sample can belong to multiple classes) as opposed to multi-class setting? Basically just as you have written in the ReadMe

docs/creating_dataset.md
```diff
@@ -298,6 +298,9 @@ SubjectID, Channel, FilePath, Label
 4, label, , 
 ```
 Note that the label field for sample 4 is left empty, this indicates that all labels are negative in Sample 4.
+In multi-label tasks, the negative class (all other ground truth classes being false for a sample) should not be
```
melanibe, line 301:
```suggestion
In multi-label tasks, the negative class (all ground truth classes being false for a sample) should not be 
```

melanibe (approved):

InnerEye/ML/dataset/scalar_dataset.py
```diff
@@ -31,57 +31,87 @@
...
+            elif label_lower in ["false", "no"]:
+                return [0.0]
+            elif label_string in ["0", "1"]:
+                return [float(label_string)]
+            else:
+                raise ValueError(f"Subject {sample_id}: Label string not recognized: '{label_string}'. "
+                             f"Should be one of true/false, yes/no or 0/1.")
+
+        if '|' in label_string or label_string.isdigit():
+            classes = [int(a) for a in label_string.split('|')]
```
dccastro, line 98: I believe this should be handled the same way as single labels above, especially casting to `float`. Possibly better to extract the parsing into a separate function.

InnerEye/ML/dataset/scalar_dataset.py
```diff
@@ -31,57 +31,87 @@
...
+
+    if is_classification_dataset:
+        if num_classes == 1:
+            label_lower = label_string.lower()
+            if label_lower in ["true", "yes"]:
+                return [1.0]
+            elif label_lower in ["false", "no"]:
+                return [0.0]
+            elif label_string in ["0", "1"]:
+                return [float(label_string)]
```
dccastro, line 92: Should we allow other values that can be cast to `float`, like `NaN`? This can be useful for when we have missing labels, for example.

ant0nsc (approved):

ant0nsc (approved):

melanibe (approved):

ant0nsc (approved):

ant0nsc (dismissed):

ant0nsc (approved):

melanibe (approved):

PR #430, opened 2021-04-09
Update dicom-rt converter

### LINKED ISSUE #429: Update DICOM-RT conversion to 1.0.1
[AB#3944](https://innereye.visualstudio.com/60ce1777-00d6-4015-82bc-488a0c00202f/_workitems/edit/3944)

### LINKED ISSUE #431: When using score.py, there should be a way of passing in a string with the model ID
All segmentations returned by an Innereye model should contain the model ID and version in the Dicom tag "software version". This must be passed into `score.py` as a commandline argument, because the code itself can't see where it was downloaded from.

javier-alvarez:
close #429 
close #431

### DIFF
--- a/InnerEye/ML/configs/segmentation/HeadAndNeckBase.py
+++ b/InnerEye/ML/configs/segmentation/HeadAndNeckBase.py
@@ -115,6 +115,7 @@ class HeadAndNeckBase(SegmentationModelBase):
             largest_connected_component_foreground_classes=ground_truth_ids,
             colours=colours,
             fill_holes=fill_holes,
+            roi_intepreted_types=["ORGAN"] * len(self.fill_holes),
             class_weights=class_weights,
             slice_exclusion_rules=slice_exclusion_rules,
             summed_probability_rules=summed_probability_rules,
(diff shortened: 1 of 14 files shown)


InnerEye/ML/configs/segmentation/HeadAndNeckBase.py
```diff
@@ -115,6 +115,7 @@ def __init__(self,
             largest_connected_component_foreground_classes=ground_truth_ids,
             colours=colours,
             fill_holes=fill_holes,
+            roi_intepreted_types=["ORGAN"] * len(self.fill_holes),
```
ant0nsc, line 118: can we use the same logic as for creating fill_holes? either use constructor arg or autopopulate?

InnerEye/ML/configs/segmentation/ProstateBase.py
```diff
@@ -55,6 +55,7 @@ def __init__(self,
             ground_truth_ids_display_names=ground_truth_ids_display_names,
             colours=colours,
             fill_holes=fill_holes,
+            roi_intepreted_types=["ORGAN"] * len(fill_holes),
```
ant0nsc, line 58: can we use the same logic as for creating fill_holes? either use constructor arg or autopopulate?

ant0nsc (changes requested):

melanibe (approved):

melanibe (approved):

ant0nsc (changes requested):
I think I did not spot a test that checks if the model_id that is passed to rt_convert is really stored in the DicomRT?

ant0nsc (approved):

melanibe (approved):

Issue #311, opened 2020-11-05
LR scheduler is not initialized correctly when recovering a run


Shruthi42:
Save and load state dicts for the learning rate scheduler so we can recover a training run correctly.

[AB#3941](https://innereye.visualstudio.com/60ce1777-00d6-4015-82bc-488a0c00202f/_workitems/edit/3941)

PR #439, opened 2021-04-19
Checkpoint recovery refactoring

melanibe:
Changes the way we handle checkpoints for run recoveries, for the following use cases:
1. If AML fails for unkown reason and we want to (manually) restart the run: provide --run_recovery_id=old_run_id tag -> this should download all checkpoints available in the old_run checkpoint folder and copy them in the new run's checkpoint folder.
2. If an AML job gets preempted, the job will automatically restart with all file system in place as interrupted -> we can find the last recovery checkpoint in the local checkpoint folder and provide it as recovery checkpoint to PL. This will avoid restarting training from scratch in case of pre-emption in the future.
3. For some use cases, we may want to keep more than one recovery checkpoint -> allow the user to specify the number of recovery checkpoint to keep.

This PR also gets rid of deprecated `start_epoch` config argument as PL will automatically restore epoch flag, optimizer, lr scheduler state from the loaded recovery checkpoint itself


melanibe:
Closes #311

InnerEye/ML/deep_learning_config.py
```diff
@@ -609,7 +611,7 @@ def get_train_epochs(self) -> List[int]:
         Returns the epochs for which training will be performed.
         :return:
         """
-        return list(range(self.start_epoch + 1, self.num_epochs + 1))
+        return list(range(self._start_epoch + 1, self.num_epochs + 1))
```
ant0nsc, line 614: I'd be tempted to kill this "get_train_epochs" thing. It's main use was to speed up training in Windows, by reducing the number of process spawns.

InnerEye/ML/common.py
```diff
@@ -85,34 +86,32 @@ def get_best_checkpoint_path(path: Path) -> Path:
...
     """
-    files_and_mod_time = [(f, f.stat().st_mtime) for f in path.glob(search_pattern)]
-    files_and_mod_time.sort(key=lambda f: f[1], reverse=True)
-    for (f, _) in files_and_mod_time[1:]:
-        logging.info(f"Removing file: {f}")
-        f.unlink()
-    if files_and_mod_time:
-        return files_and_mod_time[0][0]
-    return None
+    filenames = [f for f in path.glob(RECOVERY_CHECKPOINT_FILE_NAME + "*")]
```
ant0nsc, line 98: for testability, it would be better to split the "glob" part and the regex-part into to functions.

InnerEye/ML/common.py
```diff
@@ -85,34 +86,32 @@ def get_best_checkpoint_path(path: Path) -> Path:
...
-    Clean up all checkpoints that are found in the given folder, and keep only the "best" one. "Best" is at the moment
-    defined as being the last checkpoint, but could be based on some defined policy. The best checkpoint will be
-    renamed to `best_checkpoint.ckpt`. All other files checkpoint files
-    but the best will be removed (or an existing checkpoint renamed to be the best checkpoint).
+    Creates the best checkpoint file. "Best" is at the moment defined as being the last checkpoint, but could be
+    based on some defined policy.
+    The best checkpoint will be renamed to `best_checkpoint.ckpt`.
     :param path: The folder that contains all checkpoint files.
     """
+    logging.info(f"Files in checkpoint folder: {' '.join(p.name for p in path.glob('*'))}")
```
ant0nsc, line 114: make that logging.debug?

InnerEye/ML/deep_learning_config.py
```diff
@@ -435,6 +434,11 @@ class TrainerParams(CudaAwareConfig):
                                                            doc="Save epoch checkpoints when epoch number is a multiple "
                                                                "of recovery_checkpoint_save_interval. The intended use "
                                                                "is to allow restore training from failed runs.")
+    save_last_k_recovery_checkpoints: int = param.Integer(default=1, bounds=(-1, None),
```
ant0nsc, line 437: nitpicking: Maybe rename to recovery_checkpoints_save_last_k, so that all params to do with recovery have a shared prefix?

InnerEye/ML/model_training.py
```diff
@@ -71,6 +71,19 @@ def write_args_file(config: Any, outputs_folder: Path) -> None:
     logging.info(output)
 
 
+class InnerEyeRecoveryCheckpointCallback(ModelCheckpoint):
```
ant0nsc, line 74: class docu?

InnerEye/ML/model_training.py
```diff
@@ -71,6 +71,19 @@ def write_args_file(config: Any, outputs_folder: Path) -> None:
...
+    def __init__(self, container: LightningContainer):
+        super().__init__(dirpath=str(container.checkpoint_folder),
+                         monitor="epoch",
+                         filename=RECOVERY_CHECKPOINT_FILE_NAME + "_{epoch}",
+                         period=container.recovery_checkpoint_save_interval,
+                         save_top_k=container.save_last_k_recovery_checkpoints,
+                         mode="max")
+
+    def on_train_epoch_end(self, trainer: Trainer, pl_module: LightningModule, outputs: Any) -> None:
+        pl_module.log(name="epoch", value=trainer.current_epoch)
```
ant0nsc, line 84: Why is this happening?

InnerEye/ML/utils/checkpoint_handling.py
```diff
@@ -79,26 +80,18 @@ def additional_training_done(self) -> None:
...
+        the latest checkpoint will be present in this folder too.
         :return: Constructed checkpoint path to recover from.
         """
-        start_epoch = self.container.start_epoch
-        if start_epoch > 0 and not self.run_recovery:
-            raise ValueError("Start epoch is > 0, but no run recovery object has been provided to resume training.")
+        recovery = find_recovery_checkpoint_and_epoch(self.container.checkpoint_folder)
+        if recovery is not None:
+            local_recovery_path, recovery_epoch = recovery
+            self.container._start_epoch = recovery_epoch
```
ant0nsc, line 92: If we need that field from the outside of the class, it should not be private. recovery_start_epoch?

InnerEye/ML/utils/run_recovery.py
```diff
@@ -63,25 +63,23 @@ def download_best_checkpoints_from_child_runs(config: OutputParams, run: Run) ->
     @staticmethod
     def download_all_checkpoints_from_run(config: OutputParams, run: Run) -> RunRecovery:
         """
-        Downloads all checkpoints of the provided run: The best checkpoint and the recovery checkpoint.
-        A single folder inside the checkpoints folder will be created that contains the downloaded checkpoints.
+        Downloads all checkpoints of the provided run inside the checkpoints folder.
```
ant0nsc, line 66: This is now potentially a rather expensive operations: We are downloading top_k checkpoints + best checkpoints, rather than 2. You could probably reduce that to only getting the single "latest recovery" checkpoint - you have the logic already to get the "latest" checkpoint

Tests/ML/models/test_scalar_model.py
```diff
@@ -337,7 +346,36 @@ def test_runner1(test_output_dirs: OutputFolderForTests) -> None:
...
+    that it is a recovery run. Also checks that we update the start epoch in the config at loading time.
+    """
+    model_config = DummyClassification()
+    model_config.set_output_to(test_output_dirs.root_dir)
+    model_config.num_epochs = 6
+    model_config.recovery_checkpoint_save_interval = 2
+    model_config.save_last_k_recovery_checkpoints = -1
+    runner = MLRunner(model_config=model_config)
+    runner.setup(use_mount_or_download_dataset=False)
+    shutil.copytree(full_ml_test_data_path("recovery_dummy_checkpoints"), runner.container.checkpoint_folder)
```
ant0nsc, line 370: We completely switched away from having stored checkpoints, to simplify long-term maintainability. Better to use `create_model_and_store_checkpoint` and create the checkpoints on-the-fly.

Tests/ML/models/test_scalar_model.py
```diff
@@ -337,7 +346,36 @@ def test_runner1(test_output_dirs: OutputFolderForTests) -> None:
...
+    assert (config.checkpoint_folder / str(RECOVERY_CHECKPOINT_FILE_NAME + "_epoch=3" + CHECKPOINT_SUFFIX)).exists()
+    assert (config.checkpoint_folder / str(RECOVERY_CHECKPOINT_FILE_NAME + "_epoch=5" + CHECKPOINT_SUFFIX)).exists()
+    assert (config.checkpoint_folder / BEST_CHECKPOINT_FILE_NAME_WITH_SUFFIX).exists()
+
 
+@pytest.mark.skipif(common_util.is_windows(), reason="Has OOM issues on windows build")
+def test_runner_restart(test_output_dirs: OutputFolderForTests) -> None:
+    """
+    Test if starting training from a folder where the checkpoints folder already has recovery checkpoints picks up
+    that it is a recovery run. Also checks that we update the start epoch in the config at loading time.
```
ant0nsc, line 361: I don't see where these statements are reflected in asserts? to me, it seems this test does not provide more than test_runner1

Tests/ML/test_model_training.py
```diff
@@ -43,8 +43,6 @@
 def test_get_total_number_of_training_epochs() -> None:
     c = DeepLearningConfig(num_epochs=2, should_validate=False)
     assert c.get_total_number_of_training_epochs() == 2
-    c = DeepLearningConfig(num_epochs=10, start_epoch=5, should_validate=False)
-    assert c.get_total_number_of_training_epochs() == 5
```
ant0nsc, line 47: You could still have that assert in place, now working with the _start_epoch field that you still use.

InnerEye/ML/deep_learning_config.py
ant0nsc: OK, I can see you found a clean way of doing that, so let's just leave it in place.

ant0nsc (changes requested):

InnerEye/ML/model_training.py
melanibe: Because if we don't log epoch, we can not keep the last k checkpoints. PL lightning requires to monitor something (here epoch) when save_top_k is set to something different than None (i.e. saving more than the last).

Tests/ML/models/test_scalar_model.py
melanibe: Because here we test that we have only two checkpoints 3 and 5. If the run was restarting from epoch 0 we would have also a checkpoint for epoch=1. But i can probably make use for the `get_total_number_of_training_epochs` as you pointed out below to make that more explicit.

InnerEye/ML/common.py
melanibe: done and added two tests

InnerEye/ML/common.py
melanibe: done

InnerEye/ML/deep_learning_config.py
melanibe: yup

InnerEye/ML/utils/checkpoint_handling.py
melanibe: true

InnerEye/ML/utils/run_recovery.py
melanibe: I would rather not limit the download here, because for example for SSL logic i need to be able to tell in a custom Container argument which checkpoint to use. If we don't have to possibly to use other checkpoints it would defeat the purpose of allowing multiple checkpoints to be saved in my opinion

ant0nsc (approved):

ozan-oktay (approved):

Issue #423, opened 2021-03-18
save_outliers() fails if institutionId and seriesId undefined in dataset.csv


kh296:
The documentation on [Dataset Creation](https://github.com/microsoft/InnerEye-DeepLearning/blob/main/docs/creating_dataset.md) mentions that `institutionId` and `seriesId` are supported fields in `dataset.csv`, but doesn't give any information on how they should be defined, or on how they're used.

If `institutionId` and `seriesId` aren't included in `dataset.csv`, or are included as the headings of blank columns, InnerEye.ML.visualizers.plot_cross_validation.save_outliers() fails, with a traceback as shown below.

It would be good if:
- the use of `institutionId` and `seriesId` could be explained in the documentation;
- the code could handle `institutionId` and `seriesId` being absent from `dataset.csv`, or being the headings of blank columns.

```
2021-03-18T07:13:37Z ERROR    Model training/testing failed. Exception: list index out of range
Traceback (most recent call last):
  File "/mnt/batch/tasks/shared/LS_root/jobs/radnetinnereyedev/azureml/hd_44526ef3-837f-4cc1-98d1-750fab36518b_0/mounts/workspaceblobstore/azureml/HD_44526ef3-837f-4cc1-98d1-750fab36518b_0/InnerEye-DeepLearning/InnerEye/ML/runner.py", line 277, in run_in_situ
    self.create_ml_runner().run()
  File "/mnt/batch/tasks/shared/LS_root/jobs/radnetinnereyedev/azureml/hd_44526ef3-837f-4cc1-98d1-750fab36518b_0/mounts/workspaceblobstore/azureml/HD_44526ef3-837f-4cc1-98d1-750fab36518b_0/InnerEye-DeepLearning/InnerEye/ML/run_ml.py", line 278, in run
    self.create_ensemble_model()
  File "/mnt/batch/tasks/shared/LS_root/jobs/radnetinnereyedev/azureml/hd_44526ef3-837f-4cc1-98d1-750fab36518b_0/mounts/workspaceblobstore/azureml/HD_44526ef3-837f-4cc1-98d1-750fab36518b_0/InnerEye-DeepLearning/InnerEye/ML/run_ml.py", line 665, in create_ensemble_model
    crossval_dir = self.plot_cross_validation_and_upload_results()
[...]

PR #446, opened 2021-04-27
Guard plot_cross_validation.save_outliers for institutionId and seriesId columns confirm with unit test

dumbledad:
Fixing Issue #423 

_Overview_:

Guarding the call to `create_portal_query_for_outliers` so that the required columns are present, and providing unit test that confirms the change.

_Detail_:

The functions `save_outliers` and `create_portal_query_for_outliers` in `InnerEye.ML.cisualizers.lot_cross_validation` 
 fail if the columns _institutionId_ and _seriesId_ are missing from the data. This PR 

- Augments `create_portal_query_for_outliers` so that it checks for the required columns and raises an informative `ValueError` if they are missing. The function description is updated accordingly.
- Changes `save_outliers` in two ways. If the columns are missing
  - they are not used in the `groupby` which gleans the stats, and
  - the portal query is omitted.
- Test data is provided to check that the revised `save_outliers` produces expected data when the columns are missing
- The unit tests `test_save_outliers` and `test_create_portal_query_for_outliers` in `Tests.ML.visualizers.test_plot_cross_validation` are extended to exercise the new code with columns missing.

_Checklist_:

- [ ] ~~Run PyCharm's code cleanup tools on your Python files.~~ (I'm using VS Code and have not done this.)

### DIFF
--- a/InnerEye/ML/visualizers/plot_cross_validation.py
+++ b/InnerEye/ML/visualizers/plot_cross_validation.py
@@ -676,9 +676,11 @@ def save_outliers(config: PlotCrossValidationConfig,
 
                     f.write(f"\n\n=== METRIC: {metric_type} ===\n\n")
                     if len(outliers) > 0:
-                        outliers_summary = str(outliers.groupby(
-                            [MetricsFileColumns.Patient.value, MetricsFileColumns.Structure.value,
-                             CSV_SERIES_HEADER, CSV_INSTITUTION_HEADER])
+                        # If running inside institution there may be no CSV_SERIES_HEADER and CSV_INSTITUTION_HEADER columns
+                        groupby_columns = [MetricsFileColumns.Patient.value, MetricsFileColumns.Structure.value]
+                        if CSV_SERIES_HEADER in outliers.columns and CSV_INSTITUTION_HEADER in outliers.columns:
+                            groupby_columns += [CSV_SERIES_HEADER, CSV_INSTITUTION_HEADER]
+                        outliers_summary = str(outliers.groupby(groupby_columns)
                                                .describe()[metric_type][stats_columns]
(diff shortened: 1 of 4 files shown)


InnerEye/ML/visualizers/plot_cross_validation.py
```diff
@@ -676,14 +676,17 @@ def save_outliers(config: PlotCrossValidationConfig,
...
+                            groupby_columns += [CSV_SERIES_HEADER, CSV_INSTITUTION_HEADER]
+                        outliers_summary = str(outliers.groupby(groupby_columns)
                                                .describe()[metric_type][stats_columns]
                                                .sort_values(stats_columns, ascending=False))
                         f.write(outliers_summary)
-                        f.write("\n\n")
-                        f.write(create_portal_query_for_outliers(outliers))
+                        if CSV_INSTITUTION_HEADER in outliers.columns and CSV_SERIES_HEADER in outliers.columns:
+                            f.write("\n\n")
+                            f.write(create_portal_query_for_outliers(outliers))
```
ant0nsc, line 689: You're now checking before calling the outlier function, and inside. Can this be simplified? For example, the check could only happen inside the create_portal_query function, but if series and institution are missing, it returns an empty string - something along those lines.

ant0nsc (changes requested):

InnerEye/ML/visualizers/plot_cross_validation.py
dumbledad: I thought about that and decided that returning an empty string was not the right way to do it. Far better to make the requirement clear, in both the function description and the first few lines of code which checks the parameter for the required columns.

ant0nsc (approved):

javier-alvarez (approved):

PR #456, opened 2021-05-10
Add Covid configs

melanibe:
This PR adds configs to train Covid detection models from Chest-Xray data.


InnerEye/ML/SSL/datamodules_and_datasets/cxr_datasets.py
```diff
@@ -175,3 +176,39 @@ def _prepare_dataset(self) -> None:
...
+    """
+    Dataset class to load CovidDataset dataset as datamodule for monitoring SSL training quality directly on
+    CovidDataset data.
+    We use CVX03 against CVX12 as proxy task.
+    """
+
+    def _prepare_dataset(self) -> None:
+        self.dataset_dataframe = pd.read_csv(self.root / "dataset.csv")
+        mapping = {"0": 0, "3": 0, "1": 1, "2": 1}
+        self.dataset_dataframe["final_label"] = self.dataset_dataframe.final_label.astype(int).astype(str).apply(
```
dccastro, line 191: why do we cast to `int` and back to `str`? could we not keep it as `int` and have `mapping: Dict[int, int]` or `List[int]`? might also be worth adding a comment explaining why casting is needed (e.g. is something wrong with the values read from the file?).

InnerEye/ML/configs/classification/CovidHierarchicalModel.py
```diff
@@ -0,0 +1,241 @@
...
+    Model to train a CovidDataset model from scratch or finetune from SSL-pretrained model.
+
+    For AML you need to provide the run_id of your SSL training job as a command line argument
+    --pretraining_run_recovery_id=id_of_your_ssl_model, this will download the checkpoints of the run to your
+    machine and load the corresponding pretrained model.
+
+    To recover from a particular checkpoint from your SSL run e.g. "recovery_epoch=499.ckpt" please use hte
+    --name_of_checkpoint argument.
+    """
+    use_pretrained_model = param.Boolean(default=False, doc="TBD")
```
dccastro, line 54: are we able to document this parameter now?

InnerEye/ML/configs/classification/CovidHierarchicalModel.py
```diff
@@ -0,0 +1,241 @@
...
+                         optimizer_type=OptimizerType.Adam,
+                         num_epochs=50,
+                         l_rate_scheduler=LRSchedulerType.Step,
+                         l_rate_step_gamma=1.0,
+                         l_rate=learning_rate,
+                         l_rate_multi_step_milestones=None,
+                         **kwargs)
+        self.num_classes = 3
+        if not self.use_pretrained_model and self.freeze_encoder:
+            raise ValueError("You specified training from scratch but you asked to freeze the encoder.")
```
dccastro, line 85: may be clearer to phrase more like "No encoder to freeze when training from scratch"

InnerEye/ML/configs/classification/CovidHierarchicalModel.py
```diff
@@ -0,0 +1,241 @@
...
+    --pretraining_run_recovery_id=id_of_your_ssl_model, this will download the checkpoints of the run to your
+    machine and load the corresponding pretrained model.
+
+    To recover from a particular checkpoint from your SSL run e.g. "recovery_epoch=499.ckpt" please use hte
+    --name_of_checkpoint argument.
+    """
+    use_pretrained_model = param.Boolean(default=False, doc="TBD")
+    freeze_encoder = param.Boolean(default=False, doc="Whether to freeze the pretrained encoder or not.")
+    name_of_checkpoint = param.String(default=None, doc="Filename of checkpoint to use for recovery")
+    test_set_ids = param.String(default=None,
```
dccastro, line 57: better to indicate in the attribute name that this refers to a CSV file, e.g. `test_set_csv`, `test_ids_csv`, or `test_set_ids_csv`

InnerEye/ML/configs/classification/CovidHierarchicalModel.py
```diff
@@ -0,0 +1,241 @@
...
+                                doc="Name of the csv file in the dataset folder with the test set ids.")
```
dccastro, line 58: the doc should indicate that this CSV is expected to have at least a "series" column

InnerEye/ML/configs/classification/CovidHierarchicalModel.py
```diff
@@ -0,0 +1,241 @@
...
+        logging.info(f"name_of_checkpoint: {self.name_of_checkpoint}")
+        if not ssl_path.exists():  # for test (when it is already present) we don't need to redo this.
+            if self.name_of_checkpoint is not None:
+                path_to_checkpoint = self.extra_downloaded_run_id.checkpoints_roots[0] / self.name_of_checkpoint
+            else:
+                try:
+                    path_to_checkpoint = self.extra_downloaded_run_id.get_best_checkpoint_paths()[0]
+                    assert path_to_checkpoint.exists()
+                except AssertionError:
+                    path_to_checkpoint = self.extra_downloaded_run_id.get_recovery_checkpoint_paths()[0]
```
dccastro, line 163: not sure it's a good idea to use assertions for control flow, as they can be disabled (the one below is fine, as it's just verifying the status). any special reason not to just use a plain `if`?

dccastro (changes requested):

Shruthi42 (approved):

InnerEye/ML/SSL/datamodules_and_datasets/cxr_datasets.py
melanibe: I think everything is converted to string when we load the dataset within InnerEye actually. Maybe the conversion is not needed anymore indeed, would need to check

InnerEye/ML/configs/classification/CovidHierarchicalModel.py
melanibe: Hahah yes indeed

InnerEye/ML/configs/classification/CovidHierarchicalModel.py
melanibe: Yes good point

InnerEye/ML/configs/classification/CovidHierarchicalModel.py
melanibe: done

InnerEye/ML/SSL/datamodules_and_datasets/cxr_datasets.py
melanibe: fixed

dccastro (dismissed):

dccastro (approved):

Shruthi42 (approved):

PR #458, opened 2021-05-12
Update augmentation pipeline

melanibe:
This PR simplifies and generalizes the way we handle data augmentations:
1. The pipelining logic is now taken care of by a ImageTransformPipeline class that takes as input a list of transforms to chain together. This pipeline takes of applying transforms on 3D or 2D images. The user can choose to apply the same transformation for all channels or whether to apply different transformation for each channel (if each channel represents a different modality / time point for example). 
2. The pipeline can now work directly with out of the box torchvision transform (as long as they support [..., C, H, W] inputs). This allows to get rid of nearly all of our custom augmentations functions.
3. The conversion from pipeline of image transformation to ScalarItemAugmentation is now taken care of under the hood, the user does not need to call this wrapper for each config class. This avoids having to call with ImagePipeline wrapper and then again another wrapper. 
4. Getting rid of all the intrincated unused code for RandAugment & Co. The user has now instead complete freedom to specify the set of augmentations.


InnerEye/ML/augmentations/transform_pipeline.py
```diff
@@ -0,0 +1,168 @@
...
+            for channel in range(image.shape[1]):
+                channels.append(_convert_to_tensor_if_necessary(self.pipeline(image[:, channel, :, :].unsqueeze(1))))
+            image = torch.cat(channels, dim=1)
+        # Back to [C, Z, H, W]
+        image = torch.transpose(image, 1, 0)
+        if original_input_is_2d:
+            image = image.squeeze(1)
+        return image.to(dtype=image.dtype)
+
+    def get_scalar_item_transformation(self, item: "ScalarItem") -> "ScalarItem":
```
ant0nsc, line 90: why in quotes?

Tests/ML/augmentations/test_image_transforms.py
```diff
@@ -0,0 +1,97 @@
...
+def test_add_gaussian_noise() -> None:
+    """
+    Tests functionality of add gaussian noise
+    """
+    np.random.seed(1)
+    tensor_img = torch.ones([1, 1, 256, 256], dtype=torch.float)
+    tensor_img[..., 100:150, 100:200] = 1 / 255.
+    # Input tensor [C, Z, H, W]
+    torch.manual_seed(10)
+    transformed = AddGaussianNoise(std=0.05, p_apply=1)(tensor_img)
```
ant0nsc, line 26: maybe also add a test for p_apply=0?

Tests/ML/augmentations/test_image_transforms.py
```diff
@@ -0,0 +1,97 @@
...
+from InnerEye.ML.augmentations.image_transforms import AddGaussianNoise, ElasticTransform, ExpandChannels, RandomGamma
+
+
+def test_add_gaussian_noise() -> None:
+    """
+    Tests functionality of add gaussian noise
+    """
+    np.random.seed(1)
+    tensor_img = torch.ones([1, 1, 256, 256], dtype=torch.float)
+    tensor_img[..., 100:150, 100:200] = 1 / 255.
```
ant0nsc, line 23: here and in the code below: There are lots of repeated constants (shapes, etc), maybe reduce some of the duplication (it's a test, so I would not be too picky, but it's really a lot of duplication)

Tests/ML/augmentations/test_image_transforms.py
```diff
@@ -0,0 +1,97 @@
...
+    """
+    image = torch.ones([2, 2, 256, 256]) * 255.
+    image[..., 100:150, 100:200] = 1
+
+    # Computed expected transform
+    np.random.seed(7)
+    np.random.random(1)
+
+    shape = (256, 256)
+    dx = gaussian_filter((np.random.random(shape) * 2 - 1), 4, mode="constant", cval=0) * 34
```
ant0nsc, line 55: I'm not sure this really is the best way of writing a unit test for this transform. It effectively means that you write the code twice, but still get no confirmation if the code is doing the right thing.
Alternative suggestion for a test:
- you have a "real image" (showing something that humans can interpret) on disk
- you transform it
- you compare the result to another stored image on disk

I'm normally not a fan of stored files, but here it could help because you could quickly take a look at the transformation result and check if it makes sense.

Also, it could help to have tests for p_apply=0 and 1

Tests/ML/augmentations/test_transform_pipeline.py
```diff
@@ -0,0 +1,164 @@
...
+         ],
+        use_different_transformation_per_channel)
+
+    # Test PIL image input
+    image = PIL.Image.fromarray(image).convert("L")
+    transform(image)
+
+    # Test image as [C, H. W] tensor
+    image_as_tensor = to_tensor(image)
+    transform(image_as_tensor)
```
ant0nsc, line 47: maybe do some very basic asserts on the output? Shape? Type?

Tests/ML/augmentations/test_transform_pipeline.py
```diff
@@ -0,0 +1,164 @@
...
+    test_4d_tensor = torch.ones([25, 34, 256, 256]) * 255.
+    test_4d_tensor[..., 100:150, 100:200] = 1
+    transformed = transform(test_4d_tensor)
+    assert isinstance(transformed, torch.Tensor)
+    assert transformed.shape == torch.Size([25, 34, 224, 224])
+
+    # Same transformation should be applied to all slices and channels.
+    assert torch.isclose(transformed[0, 0], transformed[1, 1]).all() != use_different_transformation_per_channel
+
+@pytest.mark.skipif(is_windows(), reason="OOM on windows")
```
ant0nsc, line 66: maybe you can make it run also on Windows if you reduce tensor sizes? Would also make the tests a bit faster

docs/building_models.md
```diff
@@ -283,3 +283,36 @@ runs are uploaded to the parent run, in the `CrossValResults` directory. This co
 There is also a directory `BaselineComparisons`, containing the Wilcoxon test results and
 scatterplots for the ensemble, as described above for single runs.
 
+### Augmentation for classification models.
+
+For classification models, you can define an augmentation pipeline to apply to your images at training, validation
+and test time. In order, to define such a series of transformations, you will need to overload the
```
ant0nsc, line 289:
```suggestion
and test time. In order to define such a series of transformations, you will need to overload the 
```

ant0nsc (changes requested):

InnerEye/ML/augmentations/transform_pipeline.py
```diff
@@ -0,0 +1,168 @@
...
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from InnerEye.ML.dataset.scalar_sample import ScalarItem
+
+
+class ImageTransformationPipeline:
+    """
+    This class is the base class to classes built to define data augmentation transformations
+    for 3D or 2D image inputs.
```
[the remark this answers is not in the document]
melanibe: This actually works for images to. In the caller here, if the input is scalar item it applies it to the relevant attributes (img or seg) otherwise it applies it to the tensor directly. But maybe indeed we can rethink whether we can instead simply ask the user to returns a pipeline for images and one for segmentations and the data loading code makes the conversion. Will rethink it

InnerEye/ML/augmentations/transform_pipeline.py
melanibe: circular imports

InnerEye/ML/dataset/scalar_dataset.py
```diff
@@ -725,7 +725,7 @@ def load_item(self, item: ScalarDataSource) -> ScalarItem:
             center_crop_size=self.args.center_crop_size,
             image_size=self.args.image_size)
 
-        return Compose3D.apply(self.transforms, sample)
+        return Compose3D.apply(self.transforms, sample)  # type: ignore
```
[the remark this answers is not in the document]
melanibe: updated

InnerEye/ML/augmentations/transform_pipeline.py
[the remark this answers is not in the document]
melanibe: I now changed the worflow: the image pipeline now really just takes care of create a pipeline of transforms that works for 2D or 3D images (tensor or PIL.Image inputs). The conversion to ScalarItem transformation is now delegated to scalar_config. I also disantangled the transforms for image and segmentations map: we now have two functions in the scalar_config API one is "get_image_transform" that is used to set transforms to apply to images and "get_segmentations_transform" that is used to set transforms to apply to segmentations. That way for the 99,999% of cases where we only uses images input we don't need to worry about segmentation maps transforms.

InnerEye/ML/augmentations/transform_pipeline.py
```diff
@@ -0,0 +1,168 @@
...
+    The transformations are applied channel by channel, the user can specify whether to apply the same transformation
+    to each channel (no random shuffling) or whether each channel should use a different transformation (random
+    parameters of transforms shuffled for each channel).
+    """
+
+    # noinspection PyMissingConstructor
+    def __init__(self,
+                 transforms: Union[Callable, List[Callable]],
+                 use_different_transformation_per_channel: bool = False,
+                 apply_pipeline_to_segmentation_maps: bool = False):
```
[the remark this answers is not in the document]
melanibe: done, see reply above

Tests/SSL/test_ssl_containers.py
```diff
@@ -110,7 +110,7 @@ def test_load_innereye_ssl_container_cifar10_cifar100_resnet_byol() -> None:
     assert loaded_config.ssl_training_type == SSLTrainingType.BYOL
 
 
-@pytest.mark.skipif(is_windows(), reason="Too slow on windows")
+# @pytest.mark.skipif(is_windows(), reason="Too slow on windows")
```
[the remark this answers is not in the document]
melanibe: yup

Tests/ML/augmentations/test_image_transforms.py
melanibe: updated

Tests/ML/augmentations/test_image_transforms.py
melanibe: Updated, used one of the test pngs with already have in the repo

InnerEye/ML/augmentations/transform_pipeline.py
melanibe: outdated

Tests/ML/augmentations/test_image_transforms.py
melanibe: done

Tests/ML/augmentations/test_transform_pipeline.py
melanibe: updated

ant0nsc (dismissed):

InnerEye/ML/augmentations/image_transforms.py
```diff
@@ -0,0 +1,116 @@
...
+        :param: std: standard deviation of the gaussian noise to add to the image.
+        """
+        super().__init__()
+        self.p_apply = p_apply
+        self.std = std
+
+    def __call__(self, data: torch.Tensor) -> torch.Tensor:
+        if np.random.random(1) > self.p_apply:
+            return data
+        noise = torch.randn(size=data.shape[-2:]) * self.std
```
Shruthi42, line 72: Is this valid for 3D images? If not, we should add a check to ensure the images are 2D here.

InnerEye/ML/augmentations/image_transforms.py
```diff
@@ -0,0 +1,116 @@
...
+        self.p_apply = p_apply
+
+    def __call__(self, data: torch.Tensor) -> torch.Tensor:
+        if np.random.random(1) > self.p_apply:
+            return data
+        result_type = data.dtype
+        data = data.cpu().numpy()
+        shape = data.shape
+
+        dx = gaussian_filter((np.random.random(shape[-2:]) * 2 - 1), self.sigma, mode="constant", cval=0) * self.alpha
```
Shruthi42, line 108: Again, should we make sure the image is 2D here?

InnerEye/ML/augmentations/transform_pipeline.py
```diff
@@ -0,0 +1,143 @@
...
+        image = _convert_to_tensor_if_necessary(image)
+        original_input_is_2d = len(image.shape) == 3
+        # If we have a 2D image [C, H, W] expand to [Z, C, H, W]. Build-in torchvision transforms allow such 4D inputs.
+        if original_input_is_2d:
+            image = image.unsqueeze(0)
+
+        else:
+            # Some transforms assume the order of dimension is [..., C, H, W] so permute first and last dimension to
+            # obtain [Z, C, H, W]
+            image = torch.transpose(image, 1, 0)
```
Shruthi42, line 68: Can we add a check to make sure there are only 4 dimensions? Otherwise transposing index 0 and 1 will not make sense.

InnerEye/ML/augmentations/image_transforms.py
melanibe: yep, it does. All customs transforms assumes you feed an image [Z, C, H, W]

InnerEye/ML/augmentations/image_transforms.py
melanibe: nope it works for 3D too

InnerEye/ML/augmentations/transform_pipeline.py
melanibe: done

Shruthi42 (approved):

ant0nsc (approved):

Issue #459, opened 2021-05-17
Allow segmentation model inference to run on datasets without segmentations or partial segmentations


ant0nsc:
Inference should become more flexible to work also on datasets where we do not have segmentations yet, or only some of the structures segmented.

Example for Prostate model (model outputs segmentations all of bladder, prostate, rectum, femurs, etc):

1. Running on a set of CT scans where there are no segmentations present 
- No channels associated with mask or ground_truth_ids 
2. Running on CT scans where some of the segmentation are present 
- Model has ground_truth_ids = ["lung_l", "lung_r", "heart"], but dataset only has a channel for "heart" 
- Ideally a report should be generated still for all patients and structures where a channel is present in the dataset 

We may want to add a flag that triggers the current behaviour, where all of the model's ground truth ids are expected in the dataset.

Example datasets that should be considered valid after this change (for a model that segments  `["lung_l", "lung_r"]`:

no segmentations, only the CT scan is present: On such a dataset, we would not be able to compute any metrics. Inference should only output the model predictions.
```
1,ct,1/ct.nii.gz
2,ct,2/ct.nii.gz
```
Subject 1 only has `lung_l` segmented, subject 2 has all segmentation that the model knows about, Subject 3 has nothing segmented: Again, the inference run should output all predictions (`lung_l` and `lung_r`) for all 3 subjects. Report should state that `lung_l` was only present in 2 out of 3 subjects, and the average Dice score over those 2 should be computed. Report should say that `lung_r` was only present in 1 out of 3 subjects, and report average Dice (degenerate average over only 1 subject)
```
1,ct,1/ct.nii.gz
1,lung_l,1/lung_l.nii.gz
2,ct,2/ct.nii.gz
2,lung_l,2/lung_l.nii.gz
2,lung_r,2/lung_r.nii.gz
3,ct,3/ct.nii.gz
```

Changes should be made as part of the current metrics computation at inference time, and should not distinguish if training was running before or not.

[AB#4032](https://innereye.visualstudio.com/60ce1777-00d6-4015-82bc-488a0c00202f/_workitems/edit/4032)

ant0nsc:
Your main work will probably be in `evaluate_model_predictions` in `model_testing.py`

PR #465, opened 2021-05-24
Added ability to run segmentation inference module in the test data without or partial ground truth files.  

asantamariapang:
Addressing issue #459: ability to run segmentation model inference on datasets without or partial segmentation files.


InnerEye/ML/dataset/full_image_dataset.py
```diff
@@ -259,18 +261,21 @@ def _load_dataset_sources(self) -> Dict[str, PatientDatasetSource]:
                                     local_dataset_root_folder=self.args.local_dataset,
                                     image_channels=self.args.image_channels,
                                     ground_truth_channels=self.args.ground_truth_ids,
-                                    mask_channel=self.args.mask_id
+                                    mask_channel=self.args.mask_id,
+                                    for_inference=self.for_inference
```
ant0nsc, line 265: "for_inference" is not an ideal choice of name: it describes its intended use, but not what it does. I'd suggest something like "allow_incomplete_labels" or its inverse "are_all_labels_required".

InnerEye/ML/dataset/sample.py
```diff
@@ -164,8 +165,8 @@ def __post_init__(self) -> None:
         ml_util.check_size_matches(arg1=self.image, arg2=self.mask,
                                    matching_dimensions=self._get_matching_dimensions())
 
-        ml_util.check_size_matches(arg1=self.image, arg2=self.labels,
-                                   matching_dimensions=self._get_matching_dimensions())
+        # ml_util.check_size_matches(arg1=self.image, arg2=self.labels,
+        #                            matching_dimensions=self._get_matching_dimensions())
```
ant0nsc, line 169: please don't forget to fix

InnerEye/ML/utils/io_util.py
```diff
@@ -421,6 +421,10 @@ def load_labels_from_dataset_source(dataset_source: PatientDatasetSource, check_
     :param check_exclusive: Check that the labels are mutually exclusive (defaults to True)
     :return: A label sample object containing ground-truth information.
     """
+
+    if not dataset_source.ground_truth_channels:
+        return None
```
ant0nsc, line 426: the type annotation is no longer correct

Tests/ML/pipelines/test_inference.py
```diff
@@ -198,3 +207,123 @@ def shrink_dim(i: int) -> int:
 
     def get_all_child_layers(self) -> List[torch.nn.Module]:
         return list()
+
+
+def create_config_from_dataset(input_list: List, train: List, val: List, test: List) -> DummyModel:
```
ant0nsc, line 212: please complete the type annotation, it should also state what the list elements are

Tests/ML/pipelines/test_inference.py
```diff
@@ -198,3 +207,123 @@ def shrink_dim(i: int) -> int:
...
+    """
+
+    class MyDummyModel(DummyModel):
+        def get_model_train_test_dataset_splits(self, dataset_df: pd.DataFrame) -> DatasetSplits:
+            return DatasetSplits(train=dataset_df[dataset_df.subject.isin(train)],
+                                 test=dataset_df[dataset_df.subject.isin(test)],
+                                 val=dataset_df[dataset_df.subject.isin(val)])
+
+    config = MyDummyModel()
+    df = pd.DataFrame(input_list, columns=['subject', 'filePath', 'channel', 'institutionId'])
```
ant0nsc, line 225: why do we need institutionID?

InnerEye/ML/model_testing.py
```diff
@@ -231,6 +231,11 @@ def evaluate_model_predictions(process_id: int,
     """
     sample = dataset.get_samples_at_index(index=process_id)[0]
     logging.info(f"Evaluating predictions for patient {sample.patient_id}")
+
+    if sample.labels is None:
```
ant0nsc, line 235: do make sure you also handle the case when only some of the labels are present

InnerEye/ML/dataset/full_image_dataset.py
```diff
@@ -259,18 +261,21 @@ def _load_dataset_sources(self) -> Dict[str, PatientDatasetSource]:
...
                                    rows: pd.DataFrame,
                                    local_dataset_root_folder: Path,
-                                   patient_id: str) -> Tuple[List[Path], str]:
+                                   patient_id: str,
+                                   allow_incomplete_labels: bool = False) -> Tuple[List[Path], str]:
     """
     Returns: 1) The full path for files specified in the training, validation and testing datasets, and
              2) Missing channels or missing files.
 
+    :param allow_incomplete_labels: flag to enforce all ground truth labels
```
ant0nsc, line 278: for boolean args, please always specify the behaviour in both cases. Also, what happens to the output if a channels is not found?

InnerEye/ML/dataset/full_image_dataset.py
```diff
@@ -281,12 +286,15 @@ def convert_channels_to_file_paths(channels: List[str],
 
     for channel_id in channels:
         row = rows.loc[rows[CSV_CHANNEL_HEADER] == channel_id]
-        if len(row) == 0:
+        if len(row) == 0 and not allow_incomplete_labels:
             failed_channel_info += f"Patient {patient_id} does not have channel '{channel_id}'" + os.linesep
+        elif len(row) == 0 and allow_incomplete_labels:
+            # Keeps track of missing channels order
+            paths.append(Path(''))
```
ant0nsc, line 293: I feel that this is not a great choice of return value. If you return None for missing files, you can change the type annotation to Optional[Path], and already by that indicate clearly to the caller that something may be missing. Also, if you work with Path(""), the caller needs to be told clearly about that special value

InnerEye/ML/dataset/full_image_dataset.py
```diff
@@ -311,6 +320,7 @@ def load_dataset_sources(dataframe: pd.DataFrame,
     :param image_channels: The names of the image channels that should be used in the result.
     :param ground_truth_channels: The names of the ground truth channels that should be used in the result.
     :param mask_channel: The name of the mask channel that should be used in the result. This can be None.
+    :param allow_incomplete_labels: Boolean variable to allow missing ground truth files.
```
ant0nsc, line 323: describe what happens if missing.

InnerEye/ML/dataset/sample.py
```diff
@@ -132,14 +132,15 @@ class PatientDatasetSource(SampleBase):
     ground_truth_channels: List[PathOrString]
     mask_channel: Optional[PathOrString]
     metadata: PatientMetadata
+    allow_incomplete_labels: bool
```
ant0nsc, line 135: I don't see why this flag needs to be here? It would be more sensible to have ground_truth_channels as List[Optional[Path]], clearly indicating that some paths can be missing.

InnerEye/ML/metrics.py
```diff
@@ -242,14 +242,23 @@ def calculate_metrics_per_class(segmentation: np.ndarray,
                          f"the label tensor indicates that there are {number_of_classes - 1} classes.")
     binaries = binaries_from_multi_label_array(segmentation, number_of_classes)
 
-    all_classes_are_binary = [is_binary_array(ground_truth[label_id]) for label_id in range(ground_truth.shape[0])]
-    if not np.all(all_classes_are_binary):
+    binary_classes = [is_binary_array(ground_truth[label_id]) for label_id in range(ground_truth.shape[0])]
+
+    # If ground truth image is nan, then will not be used for metrics computation.
+    nan_images = [np.isnan(np.sum(ground_truth[label_id])) for label_id in range(ground_truth.shape[0])]
```
ant0nsc, line 248: np.isnan(np.sum(..)) is true if even a single entry in the ground truth ID is NaN. is that the check you wanted to implement?
Also, this is a very costly check to run. Have you considered an alternative, for example passing in a None entry in ground_truth_ids?

ant0nsc: Also, generally such special cases need to be documented clearly.

InnerEye/ML/utils/io_util.py
```diff
@@ -412,24 +412,43 @@ def load_image_in_known_formats(file: Path,
...
 
 
-def load_labels_from_dataset_source(dataset_source: PatientDatasetSource, check_exclusive: bool = True) -> np.ndarray:
+def load_labels_from_dataset_source(dataset_source: PatientDatasetSource, check_exclusive: bool = True,
+                                    mask_size: Optional[Tuple[int]] = None) -> np.ndarray:
     """
     Load labels containing segmentation binary labels in one-hot-encoding.
     In the future, this function will be used to load global class and non-imaging information as well.
 
+    :type mask_size: Image size, tuple if integers.
```
ant0nsc, line 421:
```suggestion
    :type mask_size: Image size, tuple of integers.
```

ant0nsc: Please document what mask_size is used for. Consider renaming it to image_size, because a mask can be missing.

InnerEye/ML/utils/io_util.py
```diff
@@ -412,24 +412,43 @@ def load_image_in_known_formats(file: Path,
...
+            [load_image(gt, ImageDataType.SEGMENTATION.value).image for gt in dataset_source.ground_truth_channels])
+    else:
+        assert mask_size is not None
+        label_list = []
+        for gt in dataset_source.ground_truth_channels:
+            if str(gt) == '.':
+                label_list.append(np.full(mask_size, np.NAN, ImageDataType))
+            else:
+                label_list.append(load_image(gt, ImageDataType.SEGMENTATION.value).image)
+        labels = np.stack(label_list)
```
ant0nsc, line 438: those two code paths can run out of sync. their only difference is that gt can be empty in one, but not the other case. I'd find it better to merge (effectively, take the loop solution and branch based on allow_incomplete there)

InnerEye/ML/utils/io_util.py
```diff
@@ -412,24 +412,43 @@ def load_image_in_known_formats(file: Path,
...
 
-    if check_exclusive and (sum(labels) > 1.).any():  # type: ignore
+    if not dataset_source.allow_incomplete_labels:
+        labels = np.stack(
+            [load_image(gt, ImageDataType.SEGMENTATION.value).image for gt in dataset_source.ground_truth_channels])
+    else:
+        assert mask_size is not None
+        label_list = []
+        for gt in dataset_source.ground_truth_channels:
+            if str(gt) == '.':
```
ant0nsc, line 434: This is why I'm not a fan of using Path("") as a magical value, see the above comment. Your code now effectively relies on the internals of how Path.__str__ works.

InnerEye/ML/utils/io_util.py
```diff
@@ -412,24 +412,43 @@ def load_image_in_known_formats(file: Path,
...
+                label_list.append(np.full(mask_size, np.NAN, ImageDataType))
+            else:
+                label_list.append(load_image(gt, ImageDataType.SEGMENTATION.value).image)
+        labels = np.stack(label_list)
+
+    # If ground truth image is nan, then will not be used to check check_exclusive.
+    not_nan_label_images = [labels[label_id] for label_id in range(labels.shape[0])
+                            if not np.isnan(np.sum(labels[label_id]))]
```
ant0nsc, line 442: this, to me, is another reason why huge images full of NaNs are not the best representation for missing channels.

InnerEye/ML/dataset/sample.py
ant0nsc: Or add a field to the Sample class saying which ground_truth_channels are missing?

InnerEye/ML/metrics.py
ant0nsc, line 248: thinking through this piece of code here: We are going over the HUGE tensor multiple times, quite likely going out of the range of CPU cache. It would probably be more efficient if we did all checks in the loop in line 259ff, always only looking at one of the predicted structures at a time.

InnerEye/ML/metrics.py
```diff
@@ -242,14 +242,23 @@ def calculate_metrics_per_class(segmentation: np.ndarray,
...
+
+    # Validates if not binary then nan
+    assert np.all(np.array(binary_classes) == ~np.array(nan_images))
```
ant0nsc, line 251: you can feel a list into np.all, no need to convert to np.array. This check here would be a lot easier to understand if written inside the loop below. In fact, I'm not even sure this check does the right thing. Can you add tests for these special cases?

ant0nsc: More generally, this function has gotten a lot more complicated now, please add test coverage. It only operates on numpy arrays, so should be fairly easy to write them.

InnerEye/ML/dataset/full_image_dataset.py
```diff
@@ -270,7 +270,7 @@ def convert_channels_to_file_paths(channels: List[str],
                                    rows: pd.DataFrame,
                                    local_dataset_root_folder: Path,
                                    patient_id: str,
-                                   allow_incomplete_labels: bool = False) -> Tuple[List[Path], str]:
+                                   allow_incomplete_labels: bool = False) -> Tuple[List[Optional[Path]], str]:
     """
     Returns: 1) The full path for files specified in the training, validation and testing datasets, and
              2) Missing channels or missing files.
```
ant0nsc, line 276: Please clarify what "missing channels or missing files" is. By most people's reading, I think they would expect a list

InnerEye/ML/dataset/full_image_dataset.py
```diff
@@ -290,7 +290,7 @@ def convert_channels_to_file_paths(channels: List[str],
             failed_channel_info += f"Patient {patient_id} does not have channel '{channel_id}'" + os.linesep
         elif len(row) == 0 and allow_incomplete_labels:
             # Keeps track of missing channels order
-            paths.append(Path(''))
+            paths.append(None)
```
ant0nsc, line 293: This behaviour is still not documented.

InnerEye/ML/dataset/sample.py
```diff
@@ -127,21 +127,19 @@ def get_dict(self) -> Dict[str, Any]:
 class PatientDatasetSource(SampleBase):
     """
     Dataset source locations for channels associated with a given patient in a particular dataset.
+    Please note that "ground_truth_channels" is optional.
```
ant0nsc, line 130: Please clarify. the field is not optional, but it can contain None. Please also describe what those mean.

InnerEye/ML/dataset/sample.py
```diff
@@ -154,19 +152,20 @@ class Sample(SampleBase):
     image: Union[np.ndarray, torch.Tensor]
     # (Batches if from data loader) x Z x Y x X
     mask: Union[np.ndarray, torch.Tensor]
-    # (Batches if from data loader) x Classes x Z X Y x X
-    labels: Union[np.ndarray, torch.Tensor]
+    # (Batches if from data loader) x Classes x Z X Y x X, where the first class is background
+    labels: Optional[Union[np.ndarray, torch.Tensor]]
     metadata: PatientMetadata
+    missing_labels: List[bool]
```
ant0nsc, line 158: Can you find a better variable name? indices_of_missing_labels? Also, please add a comment that describes this field in detail.

InnerEye/ML/metrics.py
```diff
@@ -223,55 +223,63 @@ def _add_zero_distances(num_segmented_surface_pixels: int, seg2ref_distance_map_
...
-    binaries = binaries_from_multi_label_array(segmentation, number_of_classes)
-
+    # For 'ground_truth', the expected C dimension is (Background Channel) + (Provided Ground Truth Channels)
+    # We can resolve the number of provided channels by subtracting the number of ground truth channels that were
+    # not provided from the number of classes
+    assert ground_truth is not None
+    assert ground_truth.shape[0] >= 2
+    num_classes_including_background = len(ground_truth_ids) + 1
+    if len(ground_truth_ids) - missing_labels.count(True) != (ground_truth.shape[0] - 1):
+        raise ValueError(f"Received {len(ground_truth_ids) - missing_labels.count(True)} foreground class names, but "
```
ant0nsc, line 251: DRY

InnerEye/ML/model_testing.py
```diff
@@ -229,23 +229,32 @@ def evaluate_model_predictions(process_id: int,
     :param results_folder: Path to results folder
     :returns [PatientMetadata, list[list]]: Patient metadata and list of computed metrics for each image.
     """
+
     sample = dataset.get_samples_at_index(index=process_id)[0]
+    assert sample.missing_labels is not None
+    if sample.labels is None:
+        logging.info(f"Ground truth label were not provided for patient {sample.patient_id}, skipping evaluation from "
+                     f"predictions")
```
ant0nsc, line 237:
```suggestion
        logging.info(f"Ground truth labels were not provided for patient {sample.patient_id}, skipping evaluation.")
```

InnerEye/ML/model_testing.py
```diff
@@ -229,23 +229,32 @@ def evaluate_model_predictions(process_id: int,
...
                                                             patient_id=sample.patient_id)
     thumbnails_folder = results_folder / THUMBNAILS_FOLDER
     thumbnails_folder.mkdir(exist_ok=True)
-    plotting.plot_contours_for_all_classes(sample,
-                                           segmentation=segmentation,
-                                           foreground_class_names=config.ground_truth_ids,
-                                           result_folder=thumbnails_folder,
-                                           image_range=config.output_range)
+    if sample.missing_labels.count(True) == 0:
+        plotting.plot_contours_for_all_classes(sample,
```
ant0nsc, line 253: Why not draw contours only for the present classes?

InnerEye/ML/utils/augmentation.py
```diff
@@ -111,7 +111,8 @@ def random_crop(sample: Sample,
     """
     Randomly crops images, mask, and labels arrays according to the crop_size argument.
     The selection of the center is dependant on background probability.
-    By default it does not center on background.
+    By default it does not center on background
```
ant0nsc, line 114:
```suggestion
    By default it does not center on background.
```

InnerEye/ML/utils/io_util.py
```diff
@@ -412,36 +412,32 @@ def load_image_in_known_formats(file: Path,
...
-        labels = np.stack(label_list)
-
-    # If ground truth image is nan, then will not be used to check check_exclusive.
-    not_nan_label_images = [labels[label_id] for label_id in range(labels.shape[0])
-                            if not np.isnan(np.sum(labels[label_id]))]
-
-    if check_exclusive and (sum(np.array(not_nan_label_images)) > 1.).any():  # type: ignore
+    :return A label sample object containing ground-truth information if channel is provided
+            If no ground-truth channels provided, label is None
+            If some ground-truth channels provided are provided, then they are loaded
```
ant0nsc, line 425:
```suggestion
            If some ground-truth channels are provided, then they are loaded
```

InnerEye/ML/utils/io_util.py
```diff
@@ -412,36 +412,32 @@ def load_image_in_known_formats(file: Path,
...
+            Background is loaded if at least one ground-truth channel is provided and is the first element of the tensor
+    :return A list of booleans indicating if ground-truth channel is missing
```
ant0nsc, line 427: Please add how long that list is

InnerEye/ML/utils/io_util.py
```diff
@@ -412,36 +412,32 @@ def load_image_in_known_formats(file: Path,
...
-            else:
-                label_list.append(load_image(gt, ImageDataType.SEGMENTATION.value).image)
-        labels = np.stack(label_list)
-
-    # If ground truth image is nan, then will not be used to check check_exclusive.
-    not_nan_label_images = [labels[label_id] for label_id in range(labels.shape[0])
-                            if not np.isnan(np.sum(labels[label_id]))]
-
-    if check_exclusive and (sum(np.array(not_nan_label_images)) > 1.).any():  # type: ignore
+    :return A label sample object containing ground-truth information if channel is provided
```
ant0nsc, line 423: Please add docu how large that tensor is

InnerEye/ML/utils/io_util.py
```diff
@@ -520,12 +516,21 @@ def load_images_from_dataset_source(dataset_source: PatientDatasetSource, check_
...
-    labels = load_labels_from_dataset_source(dataset_source, check_exclusive=check_exclusive, mask_size=mask.shape)
+    labels, missing_labels = load_labels_from_dataset_source(dataset_source, check_exclusive=check_exclusive)
+
+    # All class labels are counted, including missing or no missing ground truth class labels
+    assert len(missing_labels) == len(dataset_source.ground_truth_channels)
+    # Number of missing labels corresponds to the number of None objects in 'dataset_source.ground_truth_channels'
+    assert missing_labels.count(True) == dataset_source.ground_truth_channels.count(None)
+    # Number of not missing labels corresponds the number of provided ground truth (file name) channels.
+    assert missing_labels.count(False) == len(dataset_source.ground_truth_channels) - \
+           dataset_source.ground_truth_channels.count(None)
```
ant0nsc, line 527: Here and at other places: You tend to write a LOT of asserts. I would prefer to not litter the code with asserts that are as detailed as the ones here, in particular if all they do is check that the function just called does what the documentation says it should do. These detailed asserts are great for tests, but not in the main code.

Tests/ML/datasets/test_dataset.py
```diff
@@ -262,11 +271,17 @@ def test_cropping_dataset_has_reproducible_randomness(cropping_dataset: Cropping
...
 def test_csv_dataset_as_data_loader(normalize_fn: Any,
-                                    full_image_dataset: FullImageDataset, num_dataload_workers: int) -> None:
-    batch_size = 2
+                                    full_image_dataset: FullImageDataset, num_dataload_workers: int,
+                                    default_config: ModelConfigBase) -> None:
+
+    # Set batch size number of training classes 'default_config'
+    batch_size = len(set(default_config.get_dataset_splits().train.subject))
     # load the original images separately for comparison
+    # expected number of patients is 3
```
ant0nsc, line 280: if that's the expected number, why is that only a comment and not an assert?

InnerEye/ML/dataset/sample.py
ant0nsc, line 158: I see that you have had to change a LOT of tests because this field is mandatory. You can make it optional by saying `missing_labels: Optional[List[bool]] = None`, and recognize missing list as meaning "everything is provided".

Tests/ML/test_metrics.py
```diff
@@ -106,12 +116,16 @@ def generate_random_prediction() -> np.ndarray:
...
+    assert_metrics(random_prediction, matching_gt, missing_labels_list, 0.0)
     # check voxel spacing is being used as expected
     random_prediction2 = generate_random_prediction()
     non_matching_gt = np.stack([1 - random_prediction2, random_prediction2])
-    without_spacing = assert_metrics(random_prediction, non_matching_gt, voxel_spacing=(1, 1, 1), expected=None)
-    with_spacing = assert_metrics(random_prediction, non_matching_gt, voxel_spacing=(2.0, 2.0, 2.0), expected=None)
+    without_spacing = assert_metrics(random_prediction, non_matching_gt, missing_labels_list,
+                                     voxel_spacing=(1, 1, 1), expected=None)
+    with_spacing = assert_metrics(random_prediction, non_matching_gt, missing_labels_list,
+                                  voxel_spacing=(2.0, 2.0, 2.0), expected=None)
```
ant0nsc, line 128: Can you get your line length settings in line with the rest of the team? Such spurios changes bloat PR.

Tests/ML/visualizers/test_visualize_patches.py
```diff
@@ -37,6 +37,8 @@ def test_visualize_patch_sampling(test_output_dirs: OutputFolderForTests,
     shape = (10, 30, 30)
     foreground_classes = ["fg"]
     class_weights = equally_weighted_classes(foreground_classes)
+    # Initializes 'missing_labels_list' to 'False'
```
ant0nsc, line 40: Here and in other places: Comments that only repeat what the code does do not add much value, in my view.

Tests/ML/utils/test_augmentation.py
```diff
@@ -25,6 +25,9 @@
 valid_labels = np.zeros((number_of_classes,) + image_size)
 for c in range(number_of_classes):
     valid_labels[c, class_assignments == c] = 1
+# Since we have 5 classes and all ground truth class labels are provided, initialize 'missing_labels_list' to
+# to 'False' with  length 5
```
ant0nsc, line 29: See my other comment - please do not add comments that just repeat what the code does. If there's a special twist to why you are choosing a particular value, sure, but here I feel that's wasted space.

Tests/ML/test_metrics.py
```diff
@@ -33,21 +33,23 @@ def test_calculate_dice1() -> None:
...
     zero = np.zeros((3, 3, 3))
     one = np.ones((3, 3, 3))
+    missing_labels = [False]
 
     # ground truth is expected in one-hot encoding, but the segmentation is a map with class indices in each voxel
-    def assert_metrics(segmentation: np.ndarray, ground_truth: np.ndarray, expected_dice: float) -> None:
-        a = metrics.calculate_metrics_per_class(segmentation, ground_truth,
+    def assert_metrics(segmentation: np.ndarray, ground_truth: np.ndarray, missing_label: List[bool],
+                       expected_dice: float) -> None:
+        a = metrics.calculate_metrics_per_class(segmentation, ground_truth, missing_label,
```
ant0nsc, line 41: If you follow my suggestion to add "missing_labels" as an optional field, and recognize "None" as "nothing is missing", you can greatly simplify also all these tests here.

InnerEye/ML/dataset/full_image_dataset.py
```diff
@@ -328,16 +338,19 @@ def load_dataset_sources(dataframe: pd.DataFrame,
...
             return None
+        paths = get_paths_for_channel_ids(channels=[mask_channel])
+        if len(paths) == 0:
+            return None
         else:
-            return get_paths_for_channel_ids(channels=[mask_channel])[0]
+            return paths[0]
 
-    def get_paths_for_channel_ids(channels: List[str]) -> List[Path]:
+    def get_paths_for_channel_ids(channels: List[str]) -> List[Optional[Path]]:
```
Shruthi42, line 347: This function should take `allow_incomplete_labels` as a  parameter. Otherwise, by using the variable defined in the outer function, we will allow missing image files as well in line 352.

Shruthi42 (changes requested):

InnerEye/ML/configs/segmentation/BasicModel2Epochs.py
```diff
@@ -42,7 +42,10 @@ def __init__(self, **kwargs: Any) -> None:
             num_epochs=2,
             recovery_checkpoint_save_interval=1,
             use_mixed_precision=True,
-            azure_dataset_id=AZURE_DATASET_ID,
+            # azure_dataset_id=AZURE_DATASET_ID,
```
dumbledad, line 45: I made a copy of the dataset in `innereyepubliccheckpoint > datasets > innereyepublicdatasets/datasets/2339eba2-8ec5-4ccb-86ff-c170470ac6e2_geonorm_with_train_test_split_2020_05_26` but with all the patients other than 0 to 9 deleted (i.e. their image folders and their rows in dataset.csv deleted) because patients 0-9 are the only ones in the data splits for the model `BasicModel2EpochsMoreData`.

To make this partial ground truth I additionally removed the row `5	5/spinalcord.nii.gz	spinalcord	402ba5d42f37357f18f29af17b0846cbca4c430fea5b15a6baad5ec29dd6c9ba	b7f757fb-12e0-489e-a6da-f64895cdd229` and deleted the file `5/spinalcord.nii.gz`.

The command was 
    > python InnerEye/ML/runner.py --allow_incomplete_labels --train=False --azureml --model=BasicModel2EpochsMoreData --run_recovery_id=main_1625070510_eec68070

It is also important to include this edit of `BasicModel2Epochs.py` so that it picks up the new azure_dataset_id,
which for me was `geonorm_with_train_test_split_partial_test_data_2021_06_30`

After one bug fix ([here](https://github.com/microsoft/InnerEye-DeepLearning/blob/b6241d24b61c1590f045a4799d991bd9b0bb4370/InnerEye/ML/lightning_base.py#L165-L172)) the end-to-end test ran and the output CSVs and reports included patient counts reflecting the incomplete ground truth as expected.

.gitignore
```diff
@@ -42,6 +42,7 @@ wheels/
 .installed.cfg
 *.egg
 MANIFEST
+packages-microsoft-prod.deb
```
ant0nsc, line 45: what's that about?

InnerEye/ML/config.py
```diff
@@ -474,17 +474,24 @@ class SegmentationModelBase(ModelConfigBase):
...
     #: If true an error is raised in InnerEye.ML.utils.io_util.load_labels_from_dataset_source if the labels are not
     #: mutually exclusive. Some loss functions (e.g. SoftDice) may produce results on overlapping labels, but others (e.g.
     #: FocalLoss) will fail with a cryptic error message. Set to false if you are sure that you want to use labels that
     #: are not mutually exclusive.
     check_exclusive: bool = param.Boolean(True, doc="Raise an error if the segmentation labels are not mutually exclusive.")
 
+    allow_incomplete_labels: bool = param.Boolean(
+        default=False,
+        doc="If some test data includes patients with missing ground truth data then their data will be ignored "
+        "completely unless this flag is set. Only used for segmentation models.")
```
ant0nsc, line 493: This is implied because it lives in SegmentationModelBase

InnerEye/ML/config.py
```diff
@@ -474,17 +474,24 @@ class SegmentationModelBase(ModelConfigBase):
...
+
     #: If true an error is raised in InnerEye.ML.utils.io_util.load_labels_from_dataset_source if the labels are not
     #: mutually exclusive. Some loss functions (e.g. SoftDice) may produce results on overlapping labels, but others (e.g.
     #: FocalLoss) will fail with a cryptic error message. Set to false if you are sure that you want to use labels that
     #: are not mutually exclusive.
     check_exclusive: bool = param.Boolean(True, doc="Raise an error if the segmentation labels are not mutually exclusive.")
 
+    allow_incomplete_labels: bool = param.Boolean(
+        default=False,
+        doc="If some test data includes patients with missing ground truth data then their data will be ignored "
```
ant0nsc, line 492: I don't think we are ignoring anything?
Also, I generally advise to specify the behaviour in both True and False cases. Otherwise, people need to have lots of in-depth knowledge of the code to figure out the default behaviour.

InnerEye/ML/dataset/full_image_dataset.py
```diff
@@ -213,10 +213,11 @@ class FullImageDataset(GeneralDataset):
     """
 
     def __init__(self, args: SegmentationModelBase, data_frame: pd.DataFrame,
-                 full_image_sample_transforms: Optional[Compose3D[Sample]] = None):
+                 full_image_sample_transforms: Optional[Compose3D[Sample]] = None,
+                 allow_incomplete_labels: bool = False):
```
ant0nsc, line 217: why do we have an extra argument allow_imcomplete... here? it is already contained in args.

InnerEye/ML/lightning_base.py
```diff
@@ -161,11 +161,15 @@ def setup(self) -> None:
                 unique_ids = set(split_data[CSV_SUBJECT_HEADER])
                 for patient_id in unique_ids:
                     rows = split_data.loc[split_data[CSV_SUBJECT_HEADER] == patient_id]
+                    allow_incomplete_labels = False
+                    if isinstance(self.config, SegmentationModelBase):
+                        allow_incomplete_labels = self.config.allow_incomplete_labels  # type: ignore
```
ant0nsc, line 166: I think this is already implied by the check in line 148? If mypy complains, then replace the check in 148 with the isinstance... check

InnerEye/ML/metrics.py
```diff
@@ -56,15 +57,15 @@ class InferenceMetricsForSegmentation(InferenceMetrics):
     """
     Stores metrics for segmentation models, per execution mode and epoch.
     """
-    data_split: ModelExecutionMode
+    execution_mode: ModelExecutionMode
     metrics: float
 
     def get_metrics_log_key(self) -> str:
         """
         Gets a string name for logging the metrics specific to the execution mode (train, val, test)
```
ant0nsc, line 65: comment outdated. Also, we probably want to break down by execution mode AND data split.

InnerEye/ML/model_testing.py
```diff
@@ -126,8 +127,8 @@ def segmentation_model_test_epoch(config: SegmentationModelBase,
     where the average is taken across all non-background structures in the image.
     :param checkpoint_paths: Checkpoint paths to run inference on.
     :param config: The arguments which specify all required information.
-    :param data_split: Is the model evaluated on train, test, or validation set?
-    :param results_folder: The folder where to store the results
+    :param execution_mode: Is the model evaluated on train, test, or validation set?
```
ant0nsc, line 130: I think that's actually not correct, it is Default or Ensemble?

InnerEye/ML/plotting.py
```diff
@@ -303,6 +306,10 @@ def plot_contours_for_all_classes(sample: Sample,
         if class_index == 0:
             continue
         ground_truth = sample.labels[class_index, ...]
+
+        if np.isnan(ground_truth[0, 0, 0]):
```
ant0nsc, line 310: This check is now done at two places in the code, meaning it can run out of sync. Refactor into a method "is_missing_ground_truth"?

InnerEye/ML/utils/metrics_util.py
```diff
@@ -63,22 +63,32 @@ def to_csv(self, file_name: Path) -> None:
         del sorted_by_dice[dice_numeric]
         sorted_by_dice.to_csv(file_name, index=False, float_format=self.float_format)
 
-    def save_aggregates_to_csv(self, file_path: Path) -> None:
+    def save_aggregates_to_csv(self, file_path: Path, allow_incomplete_labels: bool = False) -> None:
```
ant0nsc, line 66: I'm a little puzzled why we need this flag here. Shouldn't it be enough to just not write metrics to the dataframe, if there is not ground truth? Everything else should fall in place automatically then? At present, the code below is rather cryptic, and I personally don't get what is going on (comments would have helped too)

InnerEye/ML/metrics.py
dumbledad: Perhaps, but all I'm doing here is making our nomenclature consistent, i.e. when we say `data_splits` we mean a list of lists of patient ids, so when we say `data_split` we should mean a list of patient ids, not an execution mode. It made it hard for me to understand the code so I am just correcting that.

.gitignore
dumbledad: To run end-to-end testing I had to install the dotnet framework on my WSL linux. Having done that and run the test this ended up in the top level directory (I guess it was downloaded there as part of the installation). I can delete it but just to be safe that the same does not happen to someone else too, I added it to the .gitignore.

InnerEye/ML/config.py
dumbledad: Fixed.

InnerEye/ML/config.py
dumbledad: Fixed.

InnerEye/ML/dataset/full_image_dataset.py
dumbledad: Good point. Fixed.

InnerEye/ML/lightning_base.py
dumbledad: Good point, I'd missed the similar check on 148. Fixed.

InnerEye/ML/plotting.py
dumbledad: That wasn't my code but yes, agreed. Fixed.

InnerEye/ML/utils/metrics_util.py
dumbledad: When `allow_incomplete_labels` is false we leave the aggregates CSV file unchanged. When it is true we add a new column, `total_patients`, so that readers can easily compare the aggregated metric's patient count with the overall number of patients. That's why we need it. I'll make that more obvious in comments

dumbledad: Done

InnerEye/ML/configs/segmentation/BasicModel2Epochs.py
dumbledad: I have also added a unit test ([here](https://github.com/microsoft/InnerEye-DeepLearning/blob/c2f0d668e8cefc1c3868de86c62131f38ca2531d/Tests/ML/test_lightning_containers.py#L286-L314)) to cover that bug.

InnerEye/ML/dataset/sample.py
dumbledad: Alberto fixed this

InnerEye/ML/metrics.py
dumbledad: Fixed

InnerEye/ML/utils/io_util.py
```diff
@@ -412,24 +412,42 @@ def load_image_in_known_formats(file: Path,
...
+        if gt is None:
+            label_list.append(np.full(image_size, np.NAN, ImageDataType))
+        else:
+            label_list.append(load_image(gt, ImageDataType.SEGMENTATION.value).image)
+    labels = np.stack(label_list)
+
+    # If ground truth image is nan, then will not be used to check check_exclusive
+    # Image is nan, if voxel at index [0, 0, 0] is NaN
+    not_nan_label_images = [labels[label_id] for label_id in range(labels.shape[0])
+                            if not np.isnan(labels[label_id][0, 0, 0])]
```
ant0nsc, line 442: You can re-use your "is_missing..." function here

Tests/ML/reports/test_segmentation_report.py
```diff
@@ -31,6 +39,43 @@ def test_generate_segmentation_report(test_output_dirs: OutputFolderForTests) ->
     contents = result_html.read_text(encoding='utf-8')
     assert 'parotid_r' in contents
 
+def _test_generate_segmentation_report_with_partial_ground_truth(test_output_dirs: OutputFolderForTests) -> None:
+    """
+    The test without partial ground truth should cover more detail, here we just check that providing
+    partial ground truth results in some labels having a lower user count.
+    """
+    reports_folder = Path(__file__).parent
```
ant0nsc, line 47: Here and above: Please don't create new custom ways of accessing files in the test folder. It reduces maintainability. Instead, use the pre-defined `full_ml_test_data_path` or something like that.

Tests/ML/test_lightning_containers.py
```diff
@@ -312,3 +313,36 @@ def test_get_hyperdrive_config(number_of_cross_validation_splits: int,
...
+            ____: str,
+            allow_incomplete_labels: bool) -> Tuple[List[Optional[Path]], str]:
+        assert allow_incomplete_labels == allow_partial_ground_truth
+        raise RuntimeError(test_done_message)
+
+    with pytest.raises(RuntimeError) as runtime_error:
+        with mock.patch("InnerEye.ML.lightning_base.convert_channels_to_file_paths") as convert_channels_to_file_paths_mock:
+            convert_channels_to_file_paths_mock.side_effect = mocked_convert_channels_to_file_paths
+            container.setup()
+    assert str(runtime_error.value) == test_done_message
```
ant0nsc, line 348: Rather than raising an exception, is it possible to use any of the mock built-in functions like `assert_called`? https://docs.python.org/3/library/unittest.mock.html

ant0nsc (changes requested):

Tests/ML/reports/test_segmentation_report.py
dumbledad: Not my code, but I'll fix that.

Tests/ML/test_lightning_containers.py
dumbledad: No, I need to stop the container.setup code, rather than needing to mock up a reasonable return value from mocked_convert_channels_to_file_paths

InnerEye/ML/utils/io_util.py
dumbledad: Good point. Done

InnerEye/ML/dataset/full_image_dataset.py
```diff
@@ -349,9 +362,11 @@ def get_paths_for_channel_ids(channels: List[str]) -> List[Path]:
         metadata = PatientMetadata.from_dataframe(dataframe, patient_id)
         dataset_sources[patient_id] = PatientDatasetSource(
             metadata=metadata,
-            image_channels=get_paths_for_channel_ids(channels=image_channels),  # type: ignore
+            image_channels=get_paths_for_channel_ids(channels=image_channels,  # type: ignore
+                                                     allow_incomplete_labels_flag=allow_incomplete_labels),
```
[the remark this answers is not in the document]
dumbledad: Fixed the image channels but ended up leaving the `# type: ignore` as an assertion didn't fix it (the types were too complicated for a simple assert: `incompatible type "List[Optional[Path]]"; expected "List[Union[Path, str]]`)

ant0nsc (approved):

dumbledad (dismissed):
Interesting, can I review a PR I am assigned to?

Shruthi42 (approved):

Shruthi42 (approved):

dumbledad (approved):

PR #466, opened 2021-05-27
Fix bug in all_azure_dataset_ids and all_dataset_mountpoints

Shruthi42:


ant0nsc (approved):

InnerEye/ML/configs/segmentation/ProstatePaper.py
```diff
@@ -11,6 +11,9 @@
 fg_classes = ["external", "femur_r", "femur_l", "rectum", "prostate", "bladder", "seminalvesicles"]
 fg_display_names = ["External", "Femur_R", "Femur_L", "Rectum", "Prostate", "Bladder", "SeminalVesicles"]
 
+# Change this string to the name of your dataset on Azure blob storage.
+AZURE_DATASET_ID = "your_dataset"
```
ant0nsc, line 15: here and above - this somehow contravenes the suggestion of subclassing those models

InnerEye/ML/deep_learning_config.py
```diff
@@ -309,18 +309,38 @@ class DatasetParams(param.Parameterized):
                        "AzureML. Use an empty string for all datasets where a randomly chosen mount/download point "
                        "should be used.")
 
+    def validate(self) -> None:
+        if not self.azure_dataset_id and self.local_dataset is None:
+            raise ValueError("Either of local_dataset or azure_dataset_id must be set.")
+
+        if not self.azure_dataset_id and len(self.extra_azure_dataset_ids) > 0:
+            raise ValueError("Please specify an azure dataset in azure_dataset_id "
+                             "before adding datasets to extra_azure_dataset_ids.")
```
ant0nsc, line 318: maybe we are being too strict here? if someone specifies both datasets in extra..., and nothing in azure_dataset, it does not change the behaviour at all? The only thing that should matter: is it consistent across dataset_ids and mount points?

Tests/Common/test_segmentation_configs.py
```diff
@@ -212,7 +186,8 @@ def test_head_and_neck_paper_with_some_ground_truth_ids(
     Check that passing a num_structures between 1 and len(defaults) generates the correct subset.
     """
     ground_truth_ids = DEFAULT_HEAD_AND_NECK_GROUND_TRUTH_IDS[:ground_truth_count]
-    config = HeadAndNeckPaper(num_structures=ground_truth_count)
+    config = HeadAndNeckPaper(local_dataset=Path("foo"),
```
ant0nsc, line 189: Do you still need that extra constructor arg "local_dataset=...", now that you hardcoded the azure dataset ID in the model? (see my comment above - I think actually this change here is better than providing a fake Azure dataset ID)

ant0nsc (approved):

InnerEye/ML/configs/segmentation/ProstatePaper.py
Shruthi42: Removed `AZURE_DATASET_ID` from these classes.

melanibe (approved):

ant0nsc (approved):

PR #482, opened 2021-06-07
Check bool parameter is either true or false (after lower)

JonathanTripp:
- Add test that a boolean command line parameter is one of: 'on', 't', 'true', 'y', 'yes', '1' (for true) or 'off', 'f', 'false', 'n', 'no', '0' (for false) after lower()
- If a flag default is False, then allow --flag to set it to True (so the ==True is unnecessary)
- If a flag default is True, then allow --no-flag to set it to False (so the ==False is unnecessary)


InnerEye/Common/generic_parsing.py
```diff
@@ -122,7 +122,12 @@ def _get_basic_type(_p: param.Parameter) -> Union[type, Callable]:
             :return: Type
             """
             if isinstance(_p, param.Boolean):
-                p_type: Union[type, Callable] = lambda x: (str(x).lower() == 'true')
+                def parse_bool(x: str) -> bool:
+                    sx = str(x).lower()
+                    if sx == 'true' or sx == 'false':
```
ant0nsc, line 127: maybe add 0 and 1 as OK values too?

JonathanTripp: I've added: y/n, yes/no, 1/0, t/f, on/off because I can imagine someone using them

ant0nsc (dismissed):

InnerEye/Common/generic_parsing.py
```diff
@@ -157,7 +162,11 @@ def list_or_dict(x: str) -> Union[Dict, List]:
             return p_type
 
         for k, p in cls.get_overridable_parameters().items():
-            parser.add_argument("--" + k, help=p.doc, type=_get_basic_type(p), default=p.default)
+            if isinstance(p, param.Boolean):
+                parser.add_argument("--" + k, help=p.doc, type=_get_basic_type(p), default=p.default,
+                                    nargs='?', const=not p.default)
```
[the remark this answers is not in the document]
JonathanTripp: Yes, I have tried to explain

InnerEye/Common/generic_parsing.py
```diff
@@ -156,8 +172,38 @@ def list_or_dict(x: str) -> Union[Dict, List]:
...
+                parser.add_argument("--" + k, help=p.doc, type=parse_bool, default=False,
+                                    nargs=argparse.OPTIONAL, const=True)
+            else:
+                # If the parameter default is True then create an exclusive group of arguments.
+                # Either --flag=value as usual
+                # Or --no-flag to store False in the parameter k.
+                group = parser.add_mutually_exclusive_group(required=False)
+                group.add_argument("--" + k, help=p.doc, type=parse_bool)
+                group.add_argument('--no-' + k, dest=k, action='store_false')
+                parser.set_defaults(**{k: p.default})
```
ant0nsc, line 198: cool stuff - could you add documentation somewhere in .md about this and all the accepted bool values?

ant0nsc (changes requested):

InnerEye/Common/generic_parsing.py
JonathanTripp: Added a section to building_models.md. Also cleaned up the other docs when they use bool arguments

ant0nsc (approved):

Shruthi42 (approved):

PR #492, opened 2021-06-15
Regression test coverage for AzureML runs

ant0nsc:
Adding the capability for any run in AzureML to compare its results against a set of baseline results.

- [x] Run PyCharm's code cleanup tools on your Python files.


InnerEye/ML/deep_learning_config.py
```diff
@@ -230,6 +230,13 @@ class WorkflowParams(param.Parameterized):
     monitoring_interval_seconds: int = param.Integer(0, doc="Seconds delay between logging GPU/CPU resource "
                                                             "statistics. If 0 or less, do not log any resource "
                                                             "statistics.")
+    regression_test_folder: Optional[Path] = \
```
dumbledad, line 233: It's odd to have test related stuff in the non-test code. I'm guessing that's essential?

ant0nsc: It is essential here, yes.

InnerEye/ML/deep_learning_config.py
```diff
@@ -230,6 +230,13 @@ class WorkflowParams(param.Parameterized):
...
+        param.ClassSelector(class_=Path, default=None, allow_None=True,
+                            doc="A path to a folder that contains a set of files. At the end of training and "
```
melanibe, line 235: Should we start the doc by "For testing purposes only"?

melanibe (approved):

InnerEye/ML/baselines_util.py
```diff
@@ -173,3 +184,119 @@ def get_comparison_baselines(outputs_folder: Path, azure_config: AzureConfig,
...
+        actual_file: Optional[Path] = None
+        if actual_folder:
+            actual_file = actual_folder / file_relative
+            if not actual_file.is_file():
+                actual_file = None
+        elif temp_folder is not None and run is not None:
+            if file_relative in files_in_run:
+                actual_file = temp_folder / file_relative
+                run.download_file(name=str(file_relative), output_file_path=str(actual_file))
+        message = compare_files(expected=file, actual=actual_file) if actual_file else "Missing file"
```
Shruthi42, line 266: This should be if actual_file.exists()

JonathanTripp (approved):

InnerEye/ML/baselines_util.py
ant0nsc: In the case of downloading from runs, run.download ensures that the file exists. In the other case, we do the check in line 260, and set it to null if it does not exist. Agreed, it's a bit twisted

InnerEye/ML/deep_learning_config.py
ant0nsc: not sure it's worth it, the field is already called "regression test"...

PR #509, opened 2021-06-24
Run inference using checkpoints from registered models

Shruthi42:
Adds the ability to run inference on registered models using the parameter `model_id`.


dumbledad (dismissed):
Checked through and all looks good. The `test_model_inference_on_single_run` test may prove useful for the PR I'm working on now!

InnerEye/Azure/azure_config.py
```diff
@@ -77,8 +77,9 @@ class AzureConfig(GenericConfig):
     pytest_mark: str = param.String(doc="If provided, run pytest instead of model training. pytest will only "
                                         "run the tests that have the mark given in this argument "
                                         "('--pytest_mark gpu' will run all tests marked with 'pytest.mark.gpu')")
-    run_recovery_id: str = param.String(doc="A run recovery id string in the form 'experiment name:run id'"
-                                            " to use for inference or recovering a model training run.")
+    run_recovery_id: str = param.String(doc="A run recovery id string in the form 'experiment name:run id' "
+                                            "to use for inference, recovering a model training run or to register "
+                                            "a model.")
```
ant0nsc, line 82: Unrelated to your PR, but why are these living in AzureConfig?

InnerEye/Azure/azure_config.py
```diff
@@ -122,6 +123,8 @@ class AzureConfig(GenericConfig):
     _workspace: Workspace = param.ClassSelector(class_=Workspace,
                                                 doc="The cached workspace object that has been created in the first"
                                                     "call to get_workspace")
+    model_id: str = param.String(doc="A model id string in the form 'model name:version' "
+                                     "to use a registered model for inference.")
```
ant0nsc, line 127: See above, this param also feels out of place here. Maybe better in WorkflowParams?

InnerEye/ML/model_testing.py
```diff
@@ -47,15 +46,15 @@
...
+               checkpoint_paths: List[Path],
                model_proc: ModelProcessing = ModelProcessing.DEFAULT) -> Optional[InferenceMetrics]:
     """
     Runs model inference on segmentation or classification models, using a given dataset (that could be training,
     test or validation set). The inference results and metrics will be stored and logged in a way that may
     differ for model categories (classification, segmentation).
     :param config: The configuration of the model
     :param data_split: Indicates which of the 3 sets (training, test, or validation) is being processed.
-    :param checkpoint_handler: Checkpoint handler object to find checkpoint paths for model initialization
+    :param checkpoint_paths: Checkpoint paths initialize model.
```
ant0nsc, line 57: incomplete sentence?

InnerEye/ML/run_ml.py
```diff
@@ -223,6 +224,16 @@ def setup(self, use_mount_or_download_dataset: bool = True) -> None:
...
 
+        if self.azure_config.pretraining_run_recovery_id is not None:
+            run_to_recover = self.azure_config.fetch_run(self.azure_config.pretraining_run_recovery_id.strip())
+            run_recovery_object = RunRecovery.download_all_checkpoints_from_run(self.container,
+                                                                                run_to_recover,
+                                                                                EXTRA_RUN_SUBFOLDER,
+                                                                                only_return_path=not is_global_rank_zero())
+            self.container.extra_downloaded_run_id = run_recovery_object
+        else:
+            self.container.extra_downloaded_run_id = None
```
ant0nsc, line 235: The field name reads a bit strange - it seems to indicate that this is a run ID, but it's a RunRecovery object (which in turn is nothing but a list of Paths).

InnerEye/ML/run_ml.py
```diff
@@ -223,6 +224,16 @@ def setup(self, use_mount_or_download_dataset: bool = True) -> None:
                                                     run_context=RUN_CONTEXT)
         self.checkpoint_handler.download_recovery_checkpoints_or_weights(only_return_path=not is_global_rank_zero())
 
+        if self.azure_config.pretraining_run_recovery_id is not None:
+            run_to_recover = self.azure_config.fetch_run(self.azure_config.pretraining_run_recovery_id.strip())
+            run_recovery_object = RunRecovery.download_all_checkpoints_from_run(self.container,
+                                                                                run_to_recover,
+                                                                                EXTRA_RUN_SUBFOLDER,
+                                                                                only_return_path=not is_global_rank_zero())
+            self.container.extra_downloaded_run_id = run_recovery_object
```
ant0nsc, line 233: This field "extra_downloaded_run_id" could do with some documentation. Also, we define that twice in DeepLearningConfig and in the container. Maybe it is better places in WorkflowParams?

InnerEye/ML/run_ml.py
```diff
@@ -483,26 +498,26 @@ def run_inference_for_lightning_models(self, checkpoint_paths: List[Path]) -> No
             if torch.distributed.is_initialized():
                 torch.distributed.destroy_process_group()
             trainer, _ = create_lightning_trainer(self.container, num_nodes=1)
+            self.container._model = type(self.container.model).load_from_checkpoint(checkpoint_path=str(checkpoint_paths[0]))
```
ant0nsc, line 501: Can we maybe make that a method on the container, so that we don't need to mess with the container's private fields?

azure-pipelines/build.yaml
```diff
@@ -49,7 +49,7 @@ steps:
   # hence don't set PYTHONPATH
   - bash: |
       source activate InnerEye
-      pytest ./Tests/ -m "not (gpu or azureml or after_training_single_run or after_training_ensemble_run or inference or after_training_2node or after_training_glaucoma_cv_run)" --doctest-modules --junitxml=junit/test-results.xml --cov=. --cov-config=.coveragerc --cov-report=xml -n 2 --dist=loadscope --verbose
+      pytest ./Tests/ -m "not (gpu or azureml or after_training_single_run or after_training_ensemble_run or inference or after_training_2node or after_training_glaucoma_cv_run or after_training_hello_container)" --doctest-modules --junitxml=junit/test-results.xml --cov=. --cov-config=.coveragerc --cov-report=xml -n 2 --dist=loadscope --verbose
```
ant0nsc, line 52: Maybe this could be done better by https://pypi.org/project/pytest-unmarked/ ? See this SO question: https://stackoverflow.com/questions/39846230/how-to-run-only-unmarked-tests-in-pytest

ant0nsc (changes requested):
Overall, looking good. Some parameters could find better homes/classes. 
Can you please add documentation around how this new flag should be used to run inference? Also, ensure that the documentation is cleared of any references to functionality that no longer exists.

azure-pipelines/build.yaml
Shruthi42: This is a bit complicated - there does not seem to be an easy way to check for custom markers only and ignore markers such as "skipif" and "parametrize".

InnerEye/Azure/azure_config.py
Shruthi42: Moved to `WorkflowParams`.

InnerEye/ML/run_ml.py
Shruthi42: I've renamed `extra_downloaded_run_id` to `pretraining_run_checkpoints` and initialized it once in WorkflowParams. I've also moved `pretraining_run_recovery_id` from AzureConfig to WorkflowParams.

InnerEye/ML/run_ml.py
Shruthi42: Renamed, see my comment below.

InnerEye/ML/run_ml.py
Shruthi42: Moving `pretraining_run_checkpoints` to WorkflowParams causes issues with initialization, I'm reverting this change for now.

dumbledad (dismissed):

dumbledad (approved):

dumbledad (approved):

InnerEye/ML/deep_learning_config.py
```diff
@@ -246,8 +256,13 @@ class WorkflowParams(param.Parameterized):
                                 "be relative to the repository root directory.")
 
     def validate(self) -> None:
-        if self.weights_url and self.local_weights_path:
-            raise ValueError("Cannot specify both local_weights_path and weights_url.")
+        if sum([bool(param) for param in [self.weights_url, self.local_weights_path, self.model_id]]) > 1:
```
ant0nsc, line 259: Not worth an extra push, but I think `any([bool(param)...])` would have been clearer

ant0nsc (approved):

InnerEye/ML/deep_learning_config.py
Shruthi42: The code throws an error if 2 or more options are set, but needs to allow the case were zero or one option is set.

PR #525, opened 2021-07-06
Enable store_dataset_sample

javier-alvarez:
storage_dataset_sample parameter was not working. This PR re-enables this optional functionality for debugging purposes.

- [x] Run PyCharm's code cleanup tools on your Python files.

### DIFF
--- a/InnerEye/ML/lightning_models.py
+++ b/InnerEye/ML/lightning_models.py
@@ -110,7 +110,8 @@ class SegmentationLightning(InnerEyeLightning):
         Computes and stores all metrics coming out of a single training step.
         :param cropped_sample: The batched image crops used for training or validation.
         :param segmentation: The segmentation that was produced by the model.
-        :param is_training: Is a training crop
+        :param is_training: If true, the method is called from `training_step`, otherwise it is called from
+        `validation_step`.
         """
         # dice_per_crop_and_class has one row per crop, with background class removed
         # Dice NaN means that both ground truth and prediction are empty.
--- a/InnerEye/ML/lightning_models.py
+++ b/InnerEye/ML/lightning_models.py
@@ -25,6 +25,7 @@ from InnerEye.ML.model_config_base import ModelConfigBase
 from InnerEye.ML.scalar_config import ScalarModelBase
 from InnerEye.ML.sequence_config import SequenceModelBase
 from InnerEye.ML.utils import image_util, metrics_util, model_util
+from InnerEye.ML.utils.dataset_util import DatasetExample, store_and_upload_example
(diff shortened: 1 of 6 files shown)


InnerEye/ML/lightning_models.py
```diff
@@ -108,6 +110,7 @@ def compute_metrics(self, cropped_sample: CroppedSample, segmentation: torch.Ten
         Computes and stores all metrics coming out of a single training step.
         :param cropped_sample: The batched image crops used for training or validation.
         :param segmentation: The segmentation that was produced by the model.
+        :param is_training: Is a training crop
```
ant0nsc, line 113: Please add descriptions for both true and false. Otherwise, people need to magically know what the other branch means.

InnerEye/ML/utils/dataset_util.py
```diff
@@ -154,8 +154,8 @@ def store_and_upload_example(dataset_example: DatasetExample,
         folder = images_folder
     else:
         folder = args.example_images_folder if args else Path("")
-    if folder != "" and not os.path.exists(folder):
-        os.mkdir(folder)
+    if folder != "":
```
ant0nsc, line 157: I'm slightly suspicious of this != "" check. Is that supposed to catch the Path("") case in the previous line? I doubt that it would, because you would end up comparing a string and a Path. 
Maybe simpler to use `folder = (args.example_folder if args else images_folder) or Path.cwd()`, then you don't need to guard the makedirs call at all.

InnerEye/ML/lightning_models.py
```diff
@@ -133,11 +136,18 @@ def compute_metrics(self, cropped_sample: CroppedSample, segmentation: torch.Ten
...
-        # if self.train_val_params.in_training_mode:
-        #     # store the sample train patch from this epoch for visualization
-        #     if batch_index == self.example_to_save and self.config.store_dataset_sample:
-        #         _store_dataset_sample(self.config, self.train_val_params.epoch, forward_pass_result,
-        #                               cropped_sample)
+
+        if is_training and self.config.store_dataset_sample:
+            # store the sample train patch from this epoch for visualization
+            # remove batches and channels
+            dataset_example = DatasetExample(image=cropped_sample.image[0][0].cpu().detach().numpy(),
```
ant0nsc, line 143: In practice, did that provide a good amount of insight? You would get a different subject in each epoch.

ant0nsc (changes requested):

ant0nsc (approved):

ant0nsc (approved):

Shruthi42 (approved):

Issue #527, opened 2021-07-06
Enable building an ensemble models from the cross validation checkpoints of a BYOL model


dumbledad:
The method `MLRunner.run_inference_for_lightning_models`  takes a list of checkpoint paths as an argument, but then makes sure that there is only one used ([here](https://github.com/microsoft/InnerEye-DeepLearning/blob/3fc71b847e9dd4ad9d0b070722558064647d95a2/InnerEye/ML/run_ml.py#L444-L445)):

```
        if len(checkpoint_paths) != 1:
            raise ValueError(f"This method expects exactly 1 checkpoint for inference, but got {len(checkpoint_paths)}")
```

We want to change this so that the checkpoints gleaned from a BYOL cross validation run can be used as an ensemble model.

- This will use the methods defined in the `InnerEyeInference` abstract class. We expect that these methods are sufficient, but an extension or redesign may be required.
- Use the simple linear regression model as the basis of an exemplar.
- We could run an end-to-end test using the UHB COVID-18 model, i.e. reproducing existing baselines and report.

dumbledad:
Is this related to #377?

PR #529, opened 2021-07-06
Enable building an ensemble model from the cross validation checkpoints of a BYO Lightning model

dumbledad:
Closing #527 

The method MLRunner.run_inference_for_lightning_models takes a list of checkpoint paths as an argument, but then makes sure that there is only one used ([here](https://github.com/microsoft/InnerEye-DeepLearning/blob/3fc71b847e9dd4ad9d0b070722558064647d95a2/InnerEye/ML/run_ml.py#L444-L445)):

```
    if len(checkpoint_paths) != 1:
        raise ValueError(f"This method expects exactly 1 checkpoint for inference, but got {len(checkpoint_paths)}")
```

We want to change this so that the checkpoints gleaned from a BYOL cross validation run can be used as an ensemble model.

- This will use the methods defined in the InnerEyeInference abstract class. We expect that these methods are sufficient, but an extension or redesign may be required.
- Use the simple linear regression model as the basis of an exemplar.
- It would be great to use another, more realistic model. The UHB Covid-19 model is not a lightning model, how about the Single Cell model?

[AB#4219](https://innereye.visualstudio.com/60ce1777-00d6-4015-82bc-488a0c00202f/_workitems/edit/4219)


ant0nsc (changes requested):
This is a good first step, but I feel the interface is not yet as easy as it should be. Anybody who wants to use ensemble models now has to define two models - and I think that's simply not going to fly. Can you think of a simpler solution?
As for testing - you have added a lot of complicated switching logic, that I was not fully able to digest. Is there a way to test that too?

dumbledad:
Replaced by #549

PR #537, opened 2021-07-13
Print warning if inference is disabled but comparison requested

JonathanTripp:
Print warning if inference is disabled but comparison requested


ant0nsc (approved):

dumbledad (approved):
Interesting mocks in the test. All looks good to me

PR #531, opened 2021-07-07
Update PL to 1.3.8

melanibe:
Update PL to 1.3.8

PR #547, opened 2021-07-27
Fix: pl_find_unused_parameters was no longer used

ant0nsc:
In #531, we accidentally removed pl_find_unused_parameters. Undoing that.

### DIFF
--- a/InnerEye/ML/model_training.py
+++ b/InnerEye/ML/model_training.py
@@ -4,18 +4,14 @@
 #  ------------------------------------------------------------------------------------------
 import logging
 import os
-import subprocess
 import sys
 from pathlib import Path
-from time import sleep
 from typing import Any, Dict, Optional, Tuple, TypeVar
 
-import numpy as np
 from pytorch_lightning import LightningModule, Trainer, seed_everything
 from pytorch_lightning.callbacks import ModelCheckpoint
 from pytorch_lightning.loggers import TensorBoardLogger
 from pytorch_lightning.plugins import DDPPlugin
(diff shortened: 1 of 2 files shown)


melanibe (approved):

melanibe (approved):

InnerEye/ML/model_training.py
```diff
@@ -131,8 +127,15 @@ def create_lightning_trainer(container: LightningContainer,
     num_gpus = container.num_gpus_per_node
     effective_num_gpus = num_gpus * num_nodes
     # Accelerator should be "ddp" when running large models in AzureML (when using DDP_spawn, we get out of GPU memory).
-    # For unit tests, only "ddp_spawn" works
-    accelerator = "ddp" if effective_num_gpus > 1 else None
+    if effective_num_gpus > 1:
+        accelerator: Optional[str] = "ddp"
+        # Initialize the DDP plugin with find_unused_parameters=False by default. If True (default), it prints out
```
vale-salvatelli, line 132: the comment is a bit confusing for me, the default seems to be indicated 2 times with opposite values. Does this should be interpreted as suggestion to set pl_find_unused_parameters=False in the container? if yes I would slightly change as below. 
```suggestion
        # Initialize the DDP plugin with find_unused_parameters=False. If True (default), it prints out
```
Also, for my own understanding, where the user is setting up this default value for the container? maybe a comment there would help too?

vale-salvatelli (approved):

JonathanTripp (approved):

PR #553, opened 2021-08-19
Fix incomplete test data module setup in Lightning inference

dccastro:
In the current version of the inference code-path for Lightning containers, we call `LightningDataModule.test_dataloader()` directly, without calling `prepare_data()` or `setup()`. This can cause crashes or unexpected behaviour if users have implemented those methods.

According to the [Lightning docs](https://pytorch-lightning.readthedocs.io/en/1.3.8/common/test_set.html), the correct way to run a test loop is to provide the full `datamodule` as an argument to `Trainer.test()`.


dumbledad (approved):

javier-alvarez:
It looks like this is breaking some tests

Shruthi42 (approved):

PR #555, opened 2021-08-24
Generalize SSL functionality to work on other datasets

vale-salvatelli:
This PR contains a few changes needed to make the SSLContainer compatible with new datasets and allow a user to run by simply creating a new augmentation config, no other changes needed to the repo, namely:

* _get_transforms has been changed to accept new datasets without the need to touch the class
* get_cxr_ssl_transform has been changed to avoid the hidden channel expansion and make that optional. It has been also renamed to get_ssl_transform because it has nothing specific to cxr
* drop_last  is now set as parameter of the InnerEyeVisionDataModule and the SSLContainer - that means it can be changed when initializing a new SSLContainer
* documentation about bringing your own SSL model has been updated


Tests/SSL/test_data_modules.py
```diff
@@ -32,8 +32,8 @@ def test_weights_innereye_module() -> None:
     """
     Tests if weights in CXR data module are correctly initialized
     """
-    transforms = get_cxr_ssl_transforms(cxr_augmentation_config,
-                                        return_two_views_per_sample=True)
+    transforms = get_ssl_transforms_from_config(cxr_augmentation_config,
```
javier-alvarez, line 35: augmentation_config? Why cxr?

InnerEye/ML/SSL/lightning_containers/ssl_container.py
```diff
@@ -223,25 +225,37 @@ def _get_transforms(self, augmentation_config: Optional[CfgNode],
         examples.
         :param dataset_name: name of the dataset, value has to be in SSLDatasetName, determines which transformation
         pipeline to return.
-        :param is_ssl_encoder_module: if True the transformation pipeline will yield two version of the image it is
-        applied on. If False, return only one transformation.
+        :param is_ssl_encoder_module: if True the transformation pipeline will yield two versions of the image it is
+        applied on and it applies the same transformations for validation. If False, return only one transformation.
```
melanibe, line 229: If think the "applies the same transformations for validation" depends on what validation transform the user defines. If the user chooses to apply the same transform at training and validation time, the pipeline would return two versions of the image. Maybe we can say something like "the pipeline will apply return two versions of the input image by applying the train (resp. validation) transformations twice. Note that if your transformation does not contain any randomness, the pipeline will return two identical copies."

Tests/SSL/test_data_modules.py
vale-salvatelli: That config has been defined for cxr, here I am importing it from Tests.SSL.test_data_modules and reusing it. It's used in other tests, that are specific to cxr, I think renaming it to something more generic could be confusing

InnerEye/ML/SSL/lightning_containers/ssl_container.py
vale-salvatelli: use_training_augmentations_for_validation=is_ssl_encoder , this is why I said it will apply the "same". I hear you on the randomness part. Changed as below, let me know if it's clear enough
"if True the transformation pipeline will yield two versions of the image it is applied on and it applies the same transformations for training and validation. Note that if your transformation does not contain any randomness, the pipeline will return two identical copies. If False, it will return only one transformation."

melanibe: Mmm i actually don't understand what you mean by "it applies the same transformations for training and validation"

melanibe (approved):

InnerEye/ML/SSL/lightning_containers/ssl_container.py
```diff
@@ -17,7 +17,7 @@
 from InnerEye.ML.SSL.datamodules_and_datasets.datamodules import CombinedDataModule, InnerEyeVisionDataModule
 from InnerEye.ML.SSL.datamodules_and_datasets.transforms_utils import InnerEyeCIFARLinearHeadTransform, \
     InnerEyeCIFARTrainTransform, \
-    get_cxr_ssl_transforms
+    get_ssl_transforms_from_config
```
ant0nsc, line 20: nitpick: linebreaks are odd on this import

InnerEye/ML/SSL/lightning_containers/ssl_container.py
```diff
@@ -223,25 +225,39 @@ def _get_transforms(self, augmentation_config: Optional[CfgNode],
         examples.
         :param dataset_name: name of the dataset, value has to be in SSLDatasetName, determines which transformation
         pipeline to return.
-        :param is_ssl_encoder_module: if True the transformation pipeline will yield two version of the image it is
-        applied on. If False, return only one transformation.
+        :param is_ssl_encoder_module: if True the transformation pipeline will yield two versions of the image it is
```
ant0nsc, line 228: I don't get the description of this argument (the original version was similarly cryptic)

ant0nsc: Also, the name of the argument is somewhat odd. It seems to refer to how the argument is used, rather than what it does. At other places, this is return_two_views_per_sample

ant0nsc (approved):
Looking good, just a few minor things

InnerEye/ML/SSL/lightning_containers/ssl_container.py
vale-salvatelli: fixed

InnerEye/ML/SSL/lightning_containers/ssl_container.py
vale-salvatelli: I completely agree on the parameter name...it took me a while to understand what this was doing. I was hoping the description could help. The problem is that this boolean is somehow used for different things and in several places. Changing the name only in this method would create confusion, we would need to change it in all the module I think

The key is that if the data module is a SSL encoder we want to return both two views per sample and use augmentations in the validation, if instead the data module is a linear head we don't want neither of the two. If you have suggestions on how to better name this parameter I'd be happy to change it

PR #558, opened 2021-09-02
Use SSLEncoder when building model in CovidModel config

Shruthi42:
Bug fix: In the CovidModel config, use an object of type `SSLEncoder` when creating a new model instead of calling `create_ssl_encoder`. Without this model weights from a finetuning run will be incompatible with the model architecture created for non-finetuning runs.


dumbledad (approved):
Another small PR, that's great! :-)

ant0nsc (approved):

PR #565, opened 2021-09-20
Fix recovery of SSL training, scale SSL training to multiple nodes

ant0nsc:
When SSL training gets interrupted on low-priority nodes, there are presently weird glitches on the metrics for the linear head. We suspect that those come from the fact that the optimizer for the linear head is not saved to the checkpoint, and hence has to re-learn all of its statistics.
This PR adds the linear head optimizer such that it is accessible to PL, and will be included in the checkpoint.

In addition, the semantics of `batch_size` in SSL training is changed: Previously it was the effective batch size, taking multiple nodes into account. This meant that the code was effectively hardcoding 16 GPUS. New behaviour: Batch size is now the batch size on a single GPU. As a consequence, we can scale to any number of GPUs without code changes.

There are also new flags `pl_limit_train_batches` and `pl_limit_val_batches` to speed up training, by reducing the number of batches processed.


InnerEye/ML/SSL/lightning_modules/byol/byol_module.py
```diff
@@ -98,10 +102,16 @@ def shared_step(self, batch: BatchType, batch_idx: int) -> T:
...
+        if optimizer_idx != 0:
+            return
         loss = self.shared_step(batch, batch_idx)
         self.log_dict({'byol/train/loss': loss, 'byol/tau': self.weight_callback.current_tau})
-        return loss
+        optimizers = self.optimizers()
+        assert len(optimizers) == 2, "Expected to get 2 optimizers, one for the embedding and one for the linear head"
+        optimizers[0].zero_grad()
+        self.manual_backward(loss)
+        optimizers[0].step()
```
ant0nsc, line 114: what about scheduler step?

InnerEye-DataQuality/InnerEyeDataQuality/deep_learning/self_supervised/simclr_module.py
```diff
@@ -2,7 +2,6 @@
 #  Copyright (c) Microsoft Corporation. All rights reserved.
```
ozan-oktay, line 2: I think it would be good to attach ssl training run results for future reference - before and after this manual optimisation change.

InnerEye/ML/deep_learning_config.py
```diff
@@ -590,6 +590,14 @@ class TrainerParams(param.Parameterized):
         param.Boolean(default=False,
                       doc="Controls the PyTorch Lightning flag 'find_unused_parameters' for the DDP plugin. "
                           "Setting it to True comes with a performance hit.")
+    pl_limit_train_batches: Optional[int] = \
```
ozan-oktay, line 593: what happens if the user specifies zero? Would it skip train/val automatically, is it tested?

ant0nsc:
superseded by #584

PR #583, opened 2021-11-11
Allow overriding AzureML experiment name

dccastro:
Currently there is no way to override the name of an AzureML experiment created with the InnerEye runner, which uses the branch name by default. This PR enables one to customise the experiment name from a `LightningContainer` or command line.

### DIFF
--- a/InnerEye/ML/deep_learning_config.py
+++ b/InnerEye/ML/deep_learning_config.py
@@ -222,6 +222,9 @@ class WorkflowParams(param.Parameterized):
     local_weights_path: List[Path] = param.List(default=[], class_=Path,
                                                 doc="A list of checkpoints paths to use for inference, "
                                                     "when the job is running outside Azure.")
+    experiment_name: str = param.String(default="",
+                                        doc="If provided, overrides the default AzureML experiment name "
+                                            "based on the current Git branch.")
     model_id: str = param.String(default="",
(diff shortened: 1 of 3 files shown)


InnerEye/ML/deep_learning_config.py
```diff
@@ -222,6 +222,9 @@ class WorkflowParams(param.Parameterized):
     local_weights_path: List[Path] = param.List(default=[], class_=Path,
                                                 doc="A list of checkpoints paths to use for inference, "
                                                     "when the job is running outside Azure.")
+    experiment_name: str = param.String(default="",
```
ant0nsc, line 225: this will mean that there are 2 fields that will auto-generate a commandline argumnet experiment_name. Will this cause problems? Can you double-check?

InnerEye/ML/deep_learning_config.py
```diff
@@ -222,6 +222,9 @@ class WorkflowParams(param.Parameterized):
...
+                                        doc="If provided, overrides the default AzureML experiment name "
+                                            "based on the current Git branch.")
```
ant0nsc, line 227: In the documentation, can you say how that interacts with the other experiment_name? (in the AzureConfig class) Also mention that in AzureConfig.experiment_name

dccastro:
Closing this in favour of the more generic solution in #589.

PR #620, opened 2021-12-14
Pre-build docker images for use in PRs

ant0nsc:
Adding this PR only for reference - #618 turned out to be a lot faster and easier to use.
This PR adds
- A build definition that builds a Docker image and pushes it to an Azure Container Registry
- Updates to all build steps to consume the docker image

What is not yet working: There's a permissions issue, the conda environment can't be updated
Also, this is suprisingly slow to consume. Consuming the docker image takes between 4 and 10min, even when using a premium tier ACR.


ant0nsc:
Closing this PR because superseded by #618

PR #621, opened 2021-12-14
Enable tiling non-PANDA WSI datasets

dccastro:
This PR implements the following major changes in the tiling/preprocessing pipeline:
- Create mask-free `LoadROId` transform using foreground auto-segmentation using [Otsu threshold](https://en.wikipedia.org/wiki/Otsu%27s_method) by default if threshold is unspecified.
- Create more generic tiling scripts (`create_tiles_dataset.py` and `azure_tiles_creation.py`).
- Update and back-up working PANDA tiling scripts as `create_panda_tiles_dataset.py` and `azure_panda_tiles_creation.py` for backward-compatibility.
- Replace OpenSlide backend with cuCIM for loading WSI files. cuCIM only works on Linux.

Additionally, I've refactored our dataset classes:
- Create `SlideKey` and `TileKey` schemas for indexing the respective batch dictionaries instead of hardcoded strings. Note that `TileKey` is not yet used in `TilesDataset` and `DeepMIL`; this will be addressed in a separate follow-up PR.
- Create base `SlidesDataset`, now inherited by the simplified `PandaDataset` and `TcgaPradDataset`.

Other:
- Add tests for slide loading, luminance, foreground seg., bounding box. Most of these run with a real `.tiff` file from the PANDA dataset, added via `git-lfs`.


.gitattributes
```diff
@@ -15,3 +15,4 @@
 *.dcm filter=lfs diff=lfs merge=lfs -text
 *.zip filter=lfs diff=lfs merge=lfs -text
 *.jpg filter=lfs diff=lfs merge=lfs -text
+*.tiff filter=lfs diff=lfs merge=lfs -text
```
vale-salvatelli, line 18: For my own knowledge: what this does?

environment.yml
```diff
@@ -20,6 +20,7 @@ dependencies:
       - azureml-tensorboard==1.36.0
       - conda-merge==0.1.5
       - cryptography==3.3.2
+      - cucim==21.10.1; platform_system=="Linux"
```
dccastro, line 23: This spec prevents the Windows builds from failing, as cuCIM is incompatible.

InnerEye/ML/Histopathology/datasets/tcga_prad_dataset.py
```diff
@@ -19,44 +20,22 @@ class TcgaPradDataset(Dataset):
     - `'image_path'` (str): absolute slide image path
     - `'label'` (int, 0 or 1): label for predicting positive or negative
     """
-    SLIDE_ID_COLUMN: str = 'slide_id'
-    CASE_ID_COLUMN: str = 'case_id'
     IMAGE_COLUMN: str = 'image_path'
     LABEL_COLUMN: str = 'label'
```
vale-salvatelli, line 24: This will not work with the current dataset, discussed separately

InnerEye/ML/Histopathology/preprocessing/create_tiles_dataset.py
```diff
@@ -217,11 +291,15 @@ def main(panda_dir: Union[str, Path], root_output_dir: Union[str, Path], level:
 
 
 if __name__ == '__main__':
-    main(panda_dir="/tmp/datasets/PANDA",
-         root_output_dir="/datadrive",
-         level=1,
+    from InnerEye.ML.Histopathology.datasets.tcga_prad_dataset import TcgaPradDataset
```
vale-salvatelli, line 294: If TcgaPrad is removed also this block should be removed - is it a problem we don't actually have a single dataset implementation that is compatible with this script

vale-salvatelli (approved):

.gitattributes
dccastro: This filter makes sure that any `.tiff` files (along with the others specified here) added to the repo will be managed by Git LFS, to avoid bloating the Git history with large binary files.

InnerEye/ML/Histopathology/preprocessing/create_tiles_dataset.py
dccastro: Following your separate suggestion, I've decided to keep TCGA-PRAD as an example, and added a clarifying comment here.

InnerEye/ML/Histopathology/datasets/tcga_prad_dataset.py
dccastro: As agreed, I'm keeping this as an example, with a clarifying comment.

vale-salvatelli: I don't see the comment on my end but I trust it has been added

Tests/ML/histopathology/datasets/test_tcga_prad_dataset.py
```diff
@@ -1,34 +0,0 @@
...
-    sample = dataset[0]
-    assert isinstance(sample, dict)
-
-    expected_keys = [dataset.SLIDE_ID_COLUMN, dataset.CASE_ID_COLUMN,
-                     dataset.IMAGE_COLUMN, dataset.LABEL_COLUMN]
-    assert all(key in sample for key in expected_keys)
-
-    image_path = sample[dataset.IMAGE_COLUMN]
-    assert isinstance(image_path, str)
-    assert os.path.isfile(image_path)
```
vale-salvatelli, line 34: To not leave things completely untested, do you think we could have a SlideDataset test? obviously we can't test the length or number of positives ... but we can test the dataset contains the expected keys and and that the content of the dict has the expected type. Looking at the dataset definition, if path is an existing path and we pass a dataset.csv, we can run these tests without need for mounting any real data. What you think?

dccastro: Now added a `test_slides_dataset.csv` and some basic tests in `test_slides_dataset.py`.

vale-salvatelli (dismissed):

maxilse (dismissed):

vale-salvatelli (approved):

mebristo (approved):

PR #622, opened 2021-12-14
Bug fix: multi-GPU jobs on a VM use wrong folders

### LINKED ISSUE #601: Multi-GPU training outside of AzureML fails when creating folder structures
Running outside azureML with >1 GPU.
Rank 0 will create a folder with a timestamp, like "20211112T...." where its files will go.
When Rank 1 starts, it will try to do exactly the same, and fail there.
What it should actually do is to write into exactly the same folder as Rank 0
We need to find a way of passing the folder name to the subsequent ranks.
Two options: commandline argument or environment variables, the latter is probably cleaner.
Possible solution:
In Rank 0, folders are created. Output folder, logs folder are stored in environment variables.
In Ranks != 0: The call to self.container.create_filesystem(self.project_root) in run_ml.py should be avoided. Instead, a DeepLearningFileSystem should be instantiated with the folders taken from environment variables.

Workaround: Do not run multi-GPU jobs on VMs. Can force that by setting `max_num_gpus=1` on the commandline.

[AB#4747](https://innereye.visualstudio.com/60ce1777-00d6-4015-82bc-488a0c00202f/_workitems/edit/4747)

ant0nsc:
Fixes #601 

- [ ] Add unit tests for all functions that you introduced or modified.
- [ ] Run PyCharm's code cleanup tools on your Python files.
- [ ] Link the correct GitHub issue for tracking.
- [ ] Update the [Changelog](CHANGELOG.md) file: Describe your change in terms of 
- [ ] When merging your PR, replace the default merge message with a description of your PR,


InnerEye/ML/deep_learning_config.py
```diff
@@ -135,8 +136,14 @@ def create(project_root: Path,
...
-            timestamp = create_unique_timestamp_id()
-            run_folder = root / f"{timestamp}_{model_name}"
+            if is_global_rank_zero():
+                timestamp = create_unique_timestamp_id()
+                run_folder = root / f"{timestamp}_{model_name}"
+            else:
+                # Handle the case where there are multiple DDP threads on the same machine outside AML.
+                # Each child process will be started with the current working directory set to be the output
+                # folder of the rank 0 process. We want all other process to write to that same folder.
+                run_folder = Path.cwd().absolute()
```
vale-salvatelli, line 146: I probably miss some knowledge about what happens elsewhere - here it would seem the output of rank 0 and of other ranks will go to different outputs folders. Wouldn't we need to save the timestamp as environment variable to share across ranks? can you help me understanding this?

ant0nsc: Rank 0 sets the current working directory to be the timestamped folder. When the jobs with rank!=0 start, they see that timestamped folder as their current working directory - so they just have to keep on writing to cwd in order to write to exactly the same folder as rank 0.

vale-salvatelli (approved):
Left a follow-up a question on Teams

Shruthi42 (approved):

PR #635, opened 2022-01-14
Panda SSL tile selection and binary evaluation

harshita-s:
This PR makes changes to `panda_tiles_dataset.py` such that tiles could be selected from the dataset based on some criteria (e.g. occupancy). It also returns a binary label (tumor/non-tumor) for each tile for online evaluation.

### DIFF
--- a/InnerEye/ML/Histopathology/datasets/panda_tiles_dataset.py
+++ b/InnerEye/ML/Histopathology/datasets/panda_tiles_dataset.py
@@ -43,7 +43,6 @@ class PandaTilesDataset(TilesDataset):
                          train=None)
         if occupancy_threshold is not None:
             dataset_df_filtered = self.dataset_df.loc[self.dataset_df['occupancy'] > occupancy_threshold]  # type: ignore
-            dataset_df_filtered.reset_index(inplace=True)
             self.dataset_df = dataset_df_filtered
 
 class PandaTilesDatasetReturnImageLabel(VisionDataset):
--- a/InnerEye/ML/Histopathology/datasets/panda_tiles_dataset.py
+++ b/InnerEye/ML/Histopathology/datasets/panda_tiles_dataset.py
@@ -54,9 +54,16 @@ class PandaTilesDatasetReturnImageLabel(VisionDataset):
                  transform: Optional[Callable] = None,
                  **kwargs: Any) -> None:
         super().__init__(root=root, transform=transform)
-        self.base_dataset = PandaTilesDataset(root=root,
+        # filter out tiles based on occupancy threshold
+        self.full_dataset = PandaTilesDataset(root=root,
                                               dataset_csv=dataset_csv,
                                               dataset_df=dataset_df)
+        self.occupancy_threshold = 0
+        dataset_df_filtered = self.full_dataset.dataset_df.loc[self.full_dataset.dataset_df['occupancy'] > self.occupancy_threshold]  # type: ignore
+        dataset_df_filtered.reset_index(inplace=True)
+        self.base_dataset = PandaTilesDataset(root=root, dataset_df=dataset_df_filtered)
+        print(len(self.full_dataset))
+        print(len(self.base_dataset))
 
     def __getitem__(self, index: int) -> Tuple:  # type: ignore
         sample = self.base_dataset[index]
@@ -64,7 +71,9 @@ class PandaTilesDatasetReturnImageLabel(VisionDataset):
         image = load_pil_image(sample[self.base_dataset.IMAGE_COLUMN])
         if self.transform:
             image = self.transform(image)
-        return image, 1
+        # get binary label
+        label = 0 if sample[self.base_dataset.LABEL_COLUMN] == 0 else 1
+        return image, label
 
     def __len__(self) -> int:
         return len(self.base_dataset)
--- a/InnerEye/ML/Histopathology/datasets/panda_tiles_dataset.py
+++ b/InnerEye/ML/Histopathology/datasets/panda_tiles_dataset.py
@@ -52,6 +52,7 @@ class PandaTilesDatasetReturnImageLabel(VisionDataset):
     class label.
     """
     occupancy_threshold = 0
+
     def __init__(self,
                  root: Path,
                  dataset_csv: Optional[Union[str, Path]] = None,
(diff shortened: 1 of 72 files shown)


InnerEye/ML/Histopathology/datasets/panda_tiles_dataset.py
```diff
@@ -54,17 +54,26 @@ def __init__(self,
                  transform: Optional[Callable] = None,
                  **kwargs: Any) -> None:
         super().__init__(root=root, transform=transform)
-        self.base_dataset = PandaTilesDataset(root=root,
+        # filter out tiles based on occupancy threshold
+        self.full_dataset = PandaTilesDataset(root=root,
                                               dataset_csv=dataset_csv,
                                               dataset_df=dataset_df)
+        self.occupancy_threshold = 0
+        dataset_df_filtered = self.full_dataset.dataset_df.loc[self.full_dataset.dataset_df['occupancy'] > self.occupancy_threshold]  # type: ignore
```
dccastro, line 62: I'd suggest to move this filtering to the `PandaTilesDataset.__init__()` with `occupancy_threshold` as an optional argument (e.g. no filtering if `None`). This way, you can simply do `self.base_dataset = PandaTilesDataset(..., occupancy_threshold=xxx)` here for SSL, but it also enables this filtering for other uses of the dataset like MIL.

InnerEye/ML/Histopathology/datasets/panda_tiles_dataset.py
```diff
@@ -54,17 +54,26 @@ def __init__(self,
...
+        dataset_df_filtered.reset_index(inplace=True)
+        self.base_dataset = PandaTilesDataset(root=root, dataset_df=dataset_df_filtered)
+        print(len(self.full_dataset))
+        print(len(self.base_dataset))
```
dccastro, line 66: As you said, it's better to remove these debugging `print`s.

InnerEye/ML/Histopathology/datasets/panda_tiles_dataset.py
harshita-s: Thanks, Daniel. Could you please explain a bit more how to include filtering in the constructor, as first I instantiate the tiles dataset to get the full dataset_df which is then filtered based on occupancy. The dataset_df is not yet assigned in the constructor.

InnerEye/ML/Histopathology/datasets/panda_tiles_dataset.py
```diff
@@ -54,17 +54,26 @@ def __init__(self,
                  transform: Optional[Callable] = None,
                  **kwargs: Any) -> None:
         super().__init__(root=root, transform=transform)
-        self.base_dataset = PandaTilesDataset(root=root,
+        # filter out tiles based on occupancy threshold
+        self.full_dataset = PandaTilesDataset(root=root,
                                               dataset_csv=dataset_csv,
                                               dataset_df=dataset_df)
+        self.occupancy_threshold = 0
```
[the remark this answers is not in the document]
vale-salvatelli: shouldn't we set this parameter as parameter of the constructor? so that it becomes something we can configure from the container? otherwise to run again this experiment we will need to change each time this class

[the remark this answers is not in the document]
dccastro: @vale-salvatelli The issue is that there's no clean way to pass arguments into this SSL dataset, because it's instantiated from by the base SimCLR container. A workaround might be to declare `occupancy_threshold` as a class attribute and overwrite it from our container.

InnerEye/ML/Histopathology/datasets/panda_tiles_dataset.py
dccastro: You can override `self.dataset_df` after `super().__init__(...)`. This way, the full dataframe is already created and you can filter it.

InnerEye/ML/Histopathology/datasets/panda_tiles_dataset.py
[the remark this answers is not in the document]
vale-salvatelli: I am in favor of this solution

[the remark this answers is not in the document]
harshita-s: I have now changed this as suggested

InnerEye/ML/Histopathology/datasets/panda_tiles_dataset.py
harshita-s: I have now made this change. Thanks for the nice idea!

InnerEye/ML/Histopathology/datasets/panda_tiles_dataset.py
harshita-s: Done, thanks

InnerEye/ML/Histopathology/datasets/panda_tiles_dataset.py
```diff
@@ -35,35 +35,36 @@ class PandaTilesDataset(TilesDataset):
...
+                 dataset_df: Optional[pd.DataFrame] = None,
+                 occupancy_threshold: Optional[float] = None) -> None:
         super().__init__(root=Path(root) / self._RELATIVE_ROOT_FOLDER,
                          dataset_csv=dataset_csv,
                          dataset_df=dataset_df,
                          train=None)
-
+        if occupancy_threshold is not None:
+            dataset_df_filtered = self.dataset_df.loc[self.dataset_df['occupancy'] > occupancy_threshold]  # type: ignore
+            dataset_df_filtered.reset_index(inplace=True)
```
dccastro, line 46: I think here we need to keep the index, because it's already been set. I'd just remove this line.

dccastro (changes requested):

InnerEye/ML/Histopathology/datasets/panda_tiles_dataset.py
harshita-s: thanks, done

dccastro (dismissed):

harshita-s:
@vale-salvatelli please let me know if you have more comments on this PR

vale-salvatelli (dismissed):

dccastro (approved):

vale-salvatelli (approved):

PR #649, opened 2022-01-31
Fix in _convert_to_tensor_if_necessary so that also np.array get converted correctly

maxilse:
Fix for the _convert_to_tensor_if_necessary method so that PIL.Image as well as np.array get converted to torch.Tensor

The PR also includes a test.

### DIFF
--- a/Tests/SSL/test_ssl_containers.py
+++ b/Tests/SSL/test_ssl_containers.py
@@ -630,24 +630,20 @@ def test_simclr_dataloader_type() -> None:
     """ This test checks if the transform pipeline of a SSL job can handle different
     data types coming from the dataloader.
     """
-    def check_types_in_batch(batch: Dict) -> None:
-        assert isinstance(batch[SSLDataModuleType.ENCODER][0][0], torch.Tensor)
-        assert isinstance(batch[SSLDataModuleType.ENCODER][0][1], torch.Tensor)
-        assert isinstance(batch[SSLDataModuleType.ENCODER][1], torch.Tensor)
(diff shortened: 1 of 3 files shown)


vale-salvatelli (approved):

vale-salvatelli:
Looks good to me! thanks for taking care of this bug

Tests/SSL/test_ssl_containers.py
```diff
@@ -622,3 +624,45 @@ def test_simclr_dataset_length(test_output_dirs: OutputFolderForTests,
...
+    """ This test checks if the transform pipeline of a SSL job can handle different
+    data types coming from the dataloader.
+    """
+    def check_types_in_batch(batch: Dict) -> None:
+        assert type(batch[SSLDataModuleType.ENCODER][0][0]) == torch.Tensor
+        assert type(batch[SSLDataModuleType.ENCODER][0][1]) == torch.Tensor
+        assert type(batch[SSLDataModuleType.ENCODER][1]) == torch.Tensor
+        assert type(batch[SSLDataModuleType.LINEAR_HEAD][0]) == torch.Tensor
+        assert type(batch[SSLDataModuleType.LINEAR_HEAD][1]) == torch.Tensor
+        assert type(batch[SSLDataModuleType.LINEAR_HEAD][2]) == torch.Tensor
```
dccastro, line 639: Here it would be safer to use `isinstance(...)`, in case we get a `Tensor` subclass, for example.

dccastro (approved):
Approved, just one suggestion for the type-checking.

javier-alvarez (approved):

vale-salvatelli (approved):

PR #650, opened 2022-02-02
Enable fine-tuning in Deepmil

harshita-s:
This PR contains the following:
- Fine-tuning can be enabled or disabled in Deepmil for PANDA.
- Correct way of initializing HistoSSL encoder.

### DIFF
--- a/InnerEye/ML/Histopathology/models/deepmil.py
+++ b/InnerEye/ML/Histopathology/models/deepmil.py
@@ -14,7 +14,7 @@ import more_itertools as mi
 
 from pytorch_lightning import LightningModule
 from torch import Tensor, argmax, mode, nn, no_grad, optim, round
-from torchmetrics import AUROC, F1, Accuracy, Precision, Recall
+from torchmetrics import AUROC, F1, Accuracy, Precision, Recall, ConfusionMatrix
 
 from InnerEye.Common import fixed_paths
 from InnerEye.ML.Histopathology.datasets.base_dataset import TilesDataset, SlidesDataset
@@ -94,6 +94,11 @@ class DeepMILModule(LightningModule):
         self.tile_size = tile_size
         self.level = level
 
+        # Metrics Objects
+        self.train_metrics = self.get_metrics()
+        self.val_metrics = self.get_metrics()
+        self.test_metrics = self.get_metrics()
+        
         self.save_hyperparameters()
 
         self.verbose = verbose
@@ -103,10 +108,6 @@ class DeepMILModule(LightningModule):
         self.loss_fn = self.get_loss()
         self.activation_fn = self.get_activation()
 
-        # Metrics Objects
-        self.train_metrics = self.get_metrics()
-        self.val_metrics = self.get_metrics()
-        self.test_metrics = self.get_metrics()
 
     def get_pooling(self) -> Tuple[Callable, int]:
         pooling_layer = self.pooling_layer(self.num_encoding,
@@ -148,7 +149,8 @@ class DeepMILModule(LightningModule):
         if self.n_classes > 1:
             return nn.ModuleDict({'accuracy': Accuracy(num_classes=self.n_classes, average='micro'),
                                   'macro_accuracy': Accuracy(num_classes=self.n_classes, average='macro'),
-                                  'weighted_accuracy': Accuracy(num_classes=self.n_classes, average='weighted')})
+                                  'weighted_accuracy': Accuracy(num_classes=self.n_classes, average='weighted'),
+                                  'confusion_matrix': ConfusionMatrix(num_classes=self.n_classes)})
         else:
             return nn.ModuleDict({'accuracy': Accuracy(),
                                    'auroc': AUROC(num_classes=self.n_classes),
@@ -162,7 +164,13 @@ class DeepMILModule(LightningModule):
         if stage not in valid_stages:
             raise Exception(f"Invalid stage. Chose one of {valid_stages}")
         for metric_name, metric_object in self.get_metrics_dict(stage).items():
-            self.log(f'{stage}/{metric_name}', metric_object, on_epoch=True, on_step=False, logger=True, sync_dist=True)
+            if metric_name == "confusion_matrix":  # and stage in ['val']:
+                #  We can't log tensors in the normal way - just print it to console
+                metric_value = metric_object.compute()
+                print(f'{stage}/{metric_name}:')
+                print(np.array(metric_value.cpu()))
+            else:
+                self.log(f'{stage}/{metric_name}', metric_object, on_epoch=True, on_step=False, logger=True, sync_dist=True)
 
     def forward(self, images: Tensor) -> Tuple[Tensor, Tensor]:  # type: ignore
         with no_grad():
--- a/InnerEye/ML/Histopathology/models/deepmil.py
+++ b/InnerEye/ML/Histopathology/models/deepmil.py
@@ -74,7 +74,7 @@ class DeepMILModule(LightningModule):
         :param tile_size: The size of each tile (default=224).
         :param level: The downsampling level (e.g. 0, 1, 2) of the tiles if available (default=1).
         :param class_names: The names of the classes if available (default=None).
-        :param is_finetune: Boolean value to enable/disable finetuning (default=False)
+        :param is_finetune: Boolean value to enable/disable finetuning (default=False).
         """
         super().__init__()
 
--- a/InnerEye/ML/Histopathology/models/deepmil.py
+++ b/InnerEye/ML/Histopathology/models/deepmil.py
@@ -173,9 +173,8 @@ class DeepMILModule(LightningModule):
                 self.log(f'{stage}/{metric_name}', metric_object, on_epoch=True, on_step=False, logger=True, sync_dist=True)
 
     def forward(self, images: Tensor) -> Tuple[Tensor, Tensor]:  # type: ignore
-        # with no_grad():
-        print(self.encoder)
-        H = self.encoder(images)                        # N X L x 1 x 1
+        with no_grad():
+            H = self.encoder(images)                        # N X L x 1 x 1
         A, M = self.aggregation_fn(H)                       # A: K x N | M: K x L
         M = M.view(-1, self.num_encoding * self.pool_out_dim)
         Y_prob = self.classifier_fn(M)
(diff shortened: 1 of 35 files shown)


ant0nsc (approved):
Code looks good - but do we have any way of testing it?

InnerEye/ML/configs/histo_configs/classification/DeepSMILEPanda.py
```diff
@@ -41,6 +42,9 @@ def __init__(self, **kwargs: Any) -> None:
             pooling_type=GatedAttentionLayer.__name__,
             # average number of tiles is 56 for PANDA
             encoding_chunk_size=60,
+            cache_mode=CacheMode.MEMORY,
+            precache_location=CacheLocation.SAME,
```
vale-salvatelli, line 46: if we're fine-tuning what is the advantage of caching(save encoded features) and pre-caching(saving the all dataset up-front)?

InnerEye/ML/configs/histo_configs/classification/DeepSMILEPanda.py
```diff
@@ -103,12 +108,14 @@ def setup(self) -> None:
...
     def get_data_module(self) -> PandaTilesDataModule:
         image_key = PandaTilesDataset.IMAGE_COLUMN
-        transform = Compose(
-            [
-                LoadTilesBatchd(image_key, progress=True),
-                EncodeTilesBatchd(image_key, self.encoder, chunk_size=self.encoding_chunk_size),
-            ]
-        )
+        if self.is_finetune:
+            transform = Compose([LoadTilesBatchd(image_key, progress=True)])
```
vale-salvatelli, line 112: should we overwrite the cache parameters in this case?

InnerEye/ML/configs/histo_configs/classification/DeepSMILEPanda.py
harshita-s: In case of finetuning, transform only includes loading tiles and not encoding the tiles. The caching will be of the tiles and not the features

Tests/SSL/test_ssl_containers.py
```diff
@@ -634,7 +634,24 @@ def test_simclr_dataloader_type() -> None:
...
+            assert isinstance(batch[1], torch.Tensor)
+            if i == 1:
+                break
+
+        for i, batch in enumerate(dataloader[SSLDataModuleType.LINEAR_HEAD]):
+            assert isinstance(batch[0], torch.Tensor)
+            assert isinstance(batch[1], torch.Tensor)
+            assert isinstance(batch[2], torch.Tensor)
+            if i == 1:
+                break
```
vale-salvatelli, line 652: I guess this is an addition from Anton or Max to avoid a separate PR :) ?

InnerEye/ML/configs/histo_configs/classification/DeepSMILEPanda.py
dccastro: @vale-salvatelli to further clarify what @harshita-s means, here it's caching the loaded tiles into memory, so they don't need to be read from disk every time. IIRC this made an epoch drop from 30min to 5min in Harshita's experiments

Tests/SSL/test_ssl_containers.py
harshita-s: yes, correct

InnerEye/ML/configs/histo_configs/classification/DeepSMILEPanda.py
```diff
@@ -41,6 +42,9 @@ def __init__(self, **kwargs: Any) -> None:
             pooling_type=GatedAttentionLayer.__name__,
             # average number of tiles is 56 for PANDA
             encoding_chunk_size=60,
+            cache_mode=CacheMode.MEMORY,
+            precache_location=CacheLocation.SAME,
+            batch_size=8,
```
vale-salvatelli, line 47: Is batch_size 8 the maximum we can handle during fine-tuning? it seems that the set of hyperparameters we will need to use for DeepMIL with fine tune and without will be pretty different. If we want to leave track of this in the form of configs file in the repo one option is to start defining some child classes, for example one for DeepSMILEPandaFrozen and one for DeepSMILEPandaTuning, where in each we overwrite the parameters of this constructor that needs to be changed

harshita-s: Thanks. I will remove this line to keep the original batch size. For me, batch size 8 worked but it will require more experimentation to determine the optimal hyperparameters for finetuning. I would not create the new class as of now, but we could revisit this when we are experimenting more with finetuning. In this PR, I just want to enable the fine-tuning and we could optimize hyperparameters in the specific experiment

InnerEye/ML/Histopathology/models/deepmil.py
```diff
@@ -187,8 +191,11 @@ def log_metrics(self,
                 log_on_epoch(self, f'{stage}/{metric_name}', metric_object)
 
     def forward(self, images: Tensor) -> Tuple[Tensor, Tensor]:  # type: ignore
-        with no_grad():
+        if self.is_finetune:
             H = self.encoder(images)                        # N X L x 1 x 1
+        else:
+            with no_grad():
```
dccastro, line 197: As a cleaner alternative to `if/else`, consider using `with torch.set_grad_enabled(self.is_finetune):` ([docs](https://pytorch.org/docs/stable/generated/torch.set_grad_enabled.html)).

InnerEye/ML/Histopathology/models/encoders.py
```diff
@@ -140,4 +140,6 @@ class HistoSSLEncoder(TileEncoder):
     def _get_encoder(self) -> Tuple[Callable, int]:
         resnet18_model = resnet18(pretrained=False)
         histossl_encoder = load_weights_to_model(self.WEIGHTS_URL, resnet18_model)
-        return setup_feature_extractor(histossl_encoder, self.input_dim)  # type: ignore
+        histossl_encoder.fc = torch.nn.Sequential()
+        num_features = 512
```
dccastro, line 144: To avoid hard-coding `num_features`, one option is to take `resnet18_model.fc.in_features`

InnerEye/ML/configs/histo_configs/classification/DeepSMILEPanda.py
```diff
@@ -103,12 +108,14 @@ def setup(self) -> None:
```
vale-salvatelli, line 108: This cuda() could be the reason why the multi-gpu option is not working - discussed in more detail at voice

vale-salvatelli: About .eval(). looking at what happens inside the forward pass of deepmil I don't think we should set it to eval mode if we're in the fine-tune case

InnerEye/ML/Histopathology/models/deepmil.py
```diff
@@ -54,7 +54,8 @@ def __init__(self,
                  slide_dataset: SlidesDataset = None,
                  tile_size: int = 224,
                  level: int = 1,
-                 class_names: Optional[List[str]] = None) -> None:
+                 class_names: Optional[List[str]] = None,
+                 is_finetune: Optional[bool] = False) -> None:
```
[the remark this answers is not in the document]
harshita-s: Thanks, changed this now

InnerEye/ML/Histopathology/models/deepmil.py
```diff
@@ -187,7 +191,7 @@ def log_metrics(self,
                 log_on_epoch(self, f'{stage}/{metric_name}', metric_object)
 
     def forward(self, images: Tensor) -> Tuple[Tensor, Tensor]:  # type: ignore
-        with no_grad():
+        with set_grad_enabled(self.is_finetune):
             H = self.encoder(images)                        # N X L x 1 x 1
```
vale-salvatelli, line 195: One additional thought: here we should call the function to encode in chunks, otherwise we will be limited in the max_bag_size as we were back at the beginning of our experiments

InnerEye/ML/configs/histo_configs/classification/DeepSMILEPanda.py
harshita-s: I have added the hyperparameters for which fine-tuning works on PANDA in the comments, as suggested earlier

InnerEye/ML/Histopathology/models/encoders.py
harshita-s: Thanks, done

InnerEye/ML/configs/histo_configs/classification/DeepSMILEPanda.py
harshita-s: This is done, thanks

InnerEye/ML/Histopathology/models/deepmil.py
harshita-s: I think doing this is more involved as monai transforms use dataset and dataloaders and it could be a separate PR.

vale-salvatelli (commented):
As discussed I think we need to make this more robust:
* makes sure it can run on multi-gpu and doesn't hit memory errors
* adding tests
But I am happy to approve as first version so that more people can help out

vale-salvatelli (approved):

InnerEye/ML/Histopathology/models/encoders.py
```diff
@@ -139,5 +139,9 @@ class HistoSSLEncoder(TileEncoder):
 
     def _get_encoder(self) -> Tuple[Callable, int]:
         resnet18_model = resnet18(pretrained=False)
+        num_features = resnet18_model.fc.in_features
         histossl_encoder = load_weights_to_model(self.WEIGHTS_URL, resnet18_model)
-        return setup_feature_extractor(histossl_encoder, self.input_dim)  # type: ignore
+        histossl_encoder.fc = torch.nn.Sequential()
+        for param in histossl_encoder.parameters():
+            param.requires_grad = False
+        return histossl_encoder, num_features  # type: ignore
```
dccastro, line 147: Minor suggestion to revert these changes in `_get_encoder()`, as `setup_feature_extractor()` now implements the same behaviour.

dccastro (approved):

vale-salvatelli (approved):

PR #656, opened 2022-02-07
[WIP] Add subsampling transform

dccastro:


maxilse (approved):

InnerEye/ML/Histopathology/models/deepmil.py
```diff
@@ -204,7 +204,7 @@ def forward(self, instances: Tensor) -> Tuple[Tensor, Tensor]:  # type: ignore
         with set_grad_enabled(self.is_finetune):
             instance_features = self.encoder(instances)                    # N X L x 1 x 1
         attentions, bag_features = self.aggregation_fn(instance_features)  # K x N | K x L
-        bag_features = bag_features.view(-1, self.num_encoding * self.pool_out_dim)
+        bag_features = bag_features.view(1, -1)
```
harshita-s, line 207: why was this changed here?

dccastro: I believe this is a more robust way to reshape the outputs, without relying on `num_encoding` and `pool_out_dim` provided by the encoder and pooling components. In particular, `MeanPoolingLayer` ignores all arguments passed to it, so this operation would have failed if we expected a different shape here.

harshita-s (approved):

PR #667, opened 2022-02-22
Automatically and linearly scale the learning rate of the SSL encoder to the number of GPUS

maxilse:
The learning rate is now linearly scaled by the number of GPUs available, e.g., lr = 0.001, 8 GPUs are available => lr=0.008.

I tested it for SimCLR and BYOL (https://ml.azure.com/experiments/id/81fa8775-1a25-47ae-9fe7-c13a6b91a421?wsid=/subscriptions/db9fc1d1-b44e-45a8-902d-8c766c255568/resourceGroups/innereyerg/providers/Microsoft.MachineLearningServices/workspaces/innereye4ws&tid=72f988bf-86f1-41af-91ab-2d7cd011db47). 

In both cases the results are as expected: 
- SimCLR: More GPUs give better representations due to larger batches == more negative eaxmples.
- BYOL: More GPUs speed up the training but do not give better results.

I did not add any test since we have these two tests already: test_simclr_num_gpus() and test_simclr_num_nodes()


InnerEye/ML/SSL/lightning_containers/ssl_container.py
```diff
@@ -147,22 +147,28 @@ def create_model(self) -> LightningModule:
         # For small images like CIFAR, if using a resnet encoder, switch the first conv layer to a 3x3 kernel instead
         # of a 7x7 conv layer.
         use_7x7_first_conv_in_resnet = False if self.ssl_training_dataset_name.value.startswith("CIFAR") else True
+
+        # Adjust the learning rate to work with multi gpu
+        gpus_per_node = self.num_gpus_per_node()
+        l_rate = self.l_rate * self.num_nodes * (gpus_per_node if gpus_per_node > 0 else 1)
+        print(f"We found {self.num_nodes * gpus_per_node} GPUs, SSL encoder learning rate has been adjusted from {self.l_rate} to {l_rate}")
```
fepegar, line 154: About this line:
1. Should it be printed only when `l_rate != self.l_rate`? Otherwise it will pollute the log with a message saying that it has been adjusted from X to X
2. Is it better to use `logging` instead?

InnerEye/ML/SSL/lightning_containers/ssl_container.py
```diff
@@ -147,22 +147,28 @@ def create_model(self) -> LightningModule:
         # For small images like CIFAR, if using a resnet encoder, switch the first conv layer to a 3x3 kernel instead
         # of a 7x7 conv layer.
         use_7x7_first_conv_in_resnet = False if self.ssl_training_dataset_name.value.startswith("CIFAR") else True
+
+        # Adjust the learning rate to work with multi gpu
```
fepegar, line 151: Should we add a comment explaining why we would want to do this? Maybe a reference to the FB paper too?

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ created.
 ## Upcoming
 
 ### Added
+- ([#667](https://github.com/microsoft/InnerEye-DeepLearning/pull/667)) Automatically and linearly scale the learning rate of the SSL encoder to the number of GPUs.
```
fepegar, line 16: Unrelated question: why do we add the link manually if GitHub does it automatically?

fepegar (commented):
LGTM in general, I just added some questions.

InnerEye/ML/SSL/lightning_containers/ssl_container.py
ant0nsc: Yes, for consistency with the rest of the codebase we should use logging.info.
Btw, your output would look funky in the corner case of running on CPU, because then it would say "We found 0 GPUs"

InnerEye/ML/SSL/lightning_containers/ssl_container.py
```diff
@@ -148,10 +148,12 @@ def create_model(self) -> LightningModule:
         # of a 7x7 conv layer.
         use_7x7_first_conv_in_resnet = False if self.ssl_training_dataset_name.value.startswith("CIFAR") else True
 
-        # Adjust the learning rate to work with multi gpu
+        # Rescale the learning rate linearly according to the number of available GPUs, as seen in: https://arxiv.org/abs/1706.02677,
+        # to avoid a drop in performance
         gpus_per_node = self.num_gpus_per_node()
-        l_rate = self.l_rate * self.num_nodes * (gpus_per_node if gpus_per_node > 0 else 1)
-        print(f"We found {self.num_nodes * gpus_per_node} GPUs, SSL encoder learning rate has been adjusted from {self.l_rate} to {l_rate}")
+        if self.num_nodes * gpus_per_node > 1:
```
fepegar, line 154: I'd precompute `self.num_nodes * gpus_per_node` into a variable, maybe called `batch_size_multiplier` or `num_gpus` or something similar, and replace the three multiplications.

fepegar (changes requested):

InnerEye/ML/SSL/lightning_containers/ssl_container.py
```diff
@@ -147,13 +147,23 @@ def create_model(self) -> LightningModule:
         # For small images like CIFAR, if using a resnet encoder, switch the first conv layer to a 3x3 kernel instead
         # of a 7x7 conv layer.
         use_7x7_first_conv_in_resnet = False if self.ssl_training_dataset_name.value.startswith("CIFAR") else True
+
+        # Rescale the learning rate linearly according to the number of available GPUs, as seen in: https://arxiv.org/abs/1706.02677,
+        # to avoid a drop in performance
+        gpus_per_node = self.num_gpus_per_node()
```
fepegar, line 153: Why is mypy complaining about this?

fepegar (dismissed):

ant0nsc (approved):

mebristo (approved):

PR #677, opened 2022-03-01
Update TorchIO version

### LINKED ISSUE #672: Update TorchIO version
A bug related to inference with patches has recently been discovered and fixed:
- https://github.com/fepegar/torchio/pull/832 

This happens only when certain conditions related to patch size, patch overlap and image size are met. I ran again the comparison shown in

- https://github.com/microsoft/InnerEye-DeepLearning/pull/666#issuecomment-1048104234

and results were exactly the same.

We should update the TorchIO version so the bug doesn't propagate to InnerEye.

[AB#5100](https://innereye.visualstudio.com/60ce1777-00d6-4015-82bc-488a0c00202f/_workitems/edit/5100)

fepegar:
Resolves #672.

The new version of TorchIO fixes a small bug discovered recently which affects patch-based inference:
- https://github.com/fepegar/torchio/pull/832


javier-alvarez (dismissed):
Can you clean the description of the PR?

fepegar:
Done. Is that better?

fepegar:
This happens because InnerEye is asking TorchIO to place a patch in a location where a larger patch is meant to be, i.e., the input size to the aggregator is smaller than the sampled patch. This happens when the model output is smaller than the input.

I have modified TorchIO so the error message when this happens is more informative.

I will submit a PR here to pad the model outputs when this situation occurs. When the model outputs are expected to be smaller than the input, a large overlap is typically specified anyway, so the patches will be cropped inside the aggregator later on, so it doesn't matter much how they are padded initially.

Hopefully, al that made some sense.

ant0nsc (dismissed):

ant0nsc (approved):

javier-alvarez (approved):

PR #678, opened 2022-03-01
Print out name of log level, not only the number

fepegar:
The logs currently show "`Setting logging level to 20`". This is not very nice.

I have added a small function that prints out "`Setting logging level to 20 (INFO)`" instead.

This is printed out using `print`. Should we use `logging.info` instead?


javier-alvarez (approved):
Update the description of the PR. Remove checklist.

javier-alvarez (approved):

ant0nsc (approved):

ant0nsc:
@fepegar , regarding print versus logging.info: Could do that as well, yes, my initial thought here was to print it irrespective of the current logging level. Any output later in the code is done via logging.info/warning.

PR #681, opened 2022-03-03
Pad model outputs if they are smaller than the inputs

fepegar:
When patches are sampled from a volume, the bounding box of the patch is saved so the inferred results can be aggregated into the original position. This requires that the sampled patch and the one that will be aggregated be of the same size.

After this pull request is merged, the model outputs will be padded so they have the same shape as the sampled patches. If we merge this into #677, the CI errors should be fixed.


ant0nsc (approved):

ant0nsc (dismissed):
Change looks good - but when are we running into that problem? If it's easily doable, can there be a test for that?

fepegar:
> when are we running into that problem?

When the output of a model is smaller than the input. For example, when padding is not used in a U-Net (as in the original implementation):

U-Net

> can there be a test for that?

I think there's already one (with a typo in the name):

https://github.com/microsoft/InnerEye-DeepLearning/blob/5336a67f66f311e50149469e4e5c6ed6a5ddc280/Tests/ML/pipelines/test_inference.py#L39-L43

I don't know why [the test is passing on Azure pipelines](https://innereye.visualstudio.com/InnerEye/_build/results?buildId=47472&view=logs&j=3afc50db-e620-5b81-6016-870a6976ad29&t=0fccc816-fae5-5da6-b270-57c45cc94230)... It fails on the main branch on my machine.

ant0nsc (approved):

javier-alvarez (approved):

PR #682, opened 2022-03-03
Ensure the shape of input patches is compatible with model constraints

fepegar:
This change uses TorchIO's [`EnsureShapeMultiple`](https://torchio.readthedocs.io/transforms/preprocessing.html#torchio.transforms.EnsureShapeMultiple) to ensure that the inputs are a multiple of certain number which is specified in the model size constraints.


Shruthi42 (approved):

Tests/ML/pipelines/test_inference_smallimages.py
```diff
@@ -59,15 +59,6 @@ def run_inference_on_unet(size: TupleInt3) -> None:
         image_util.check_array_range(p)
 
 
-def test_inference_on_too_small_image() -> None:
-    """
```
javier-alvarez, line 63: Why delete a test?

fepegar: Because images will never be too small after merging
- #681

javier-alvarez: The other PR does not have tests. Is it not possible to have a test to demonstrate that this is correct?

javier-alvarez (approved):

Tests/ML/pipelines/test_inference_smallimages.py
fepegar: I believe all this is covered by the tests in https://github.com/microsoft/InnerEye-DeepLearning/blob/main/Tests/ML/models/architectures/test_crop_size_multiple.py

ant0nsc (approved):

PR #720, opened 2022-04-26
DOC: Add Read the Docs config file

### LINKED ISSUE #706: Documentation hosting
Would we like to have the documentation on Read the Docs? I'm happy to work on that if we would.

fepegar:
The Read the Docs (RTD) build is failing as requirements cannot be imported: https://readthedocs.org/projects/inereye-deeplearning/builds/16688167/

In this PR, an RTD config file is added that specifies the path to the requirements file needed for building.

If it works, it will resolve #706.


fepegar:
The requirements added in b8af514 are for the documentation itself, but we still need to tell RTD where to look for the requirements needed for the package itself. We can probably get inspiration from https://github.com/microsoft/hi-ml.

PR #750, opened 2022-06-21
ENH: Remove cucim + change lock script line endings

### LINKED ISSUE #749: Environment definition contains cucim, but we don't need/use it
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Bug summary

The default Linux environment file contains the `cucim` package. To the best of my knowledge, this is not required for any of the InnerEye functionality. `Cucim` is only used in the hi-ml-histopathology module, which we don't really use in InnerEye.
This leads to problems as reported by @kh296 in https://github.com/microsoft/InnerEye-DeepLearning/issues/741

### Code for reproduction

```Python
See https://github.com/microsoft/InnerEye-DeepLearning/issues/741
```

### Actual outcome

n/a

### Error messages

_No response_

### Expected outcome

Creating a Conda environment on a non-GPU machine should work just fine.

### System info

_No response_

[AB#6272](https://innereye.visualstudio.com/60ce1777-00d6-4015-82bc-488a0c00202f/_workitems/edit/6272)

peterhessey:
Closes #749 

Removed `cucim` and locked secondary dependencies accordingly. Also changes lock script line endings, they seem to have been set to CRLF at some point, switched back to LF now so that it can be run from a bash shell.

### DIFF
--- a/TestSubmodule/environment.yml
+++ b/TestSubmodule/environment.yml
@@ -16,7 +16,7 @@ dependencies:
   - ffmpeg=4.2.2=h20bf706_0
   - freetype=2.11.0=h70c0345_0
   - giflib=5.2.1=h7b6447c_0
-  - gmp=6.2.1=h2531618_2
+  - gmp=6.2.1=h295c915_3
   - gnutls=3.6.15=he1e5248_0
   - intel-openmp=2021.4.0=h06a4308_3561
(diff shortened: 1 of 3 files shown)


ant0nsc (approved):

fepegar (approved):
Thanks!

PR #752, opened 2022-06-22
ENH: Move hi-ml and fastMRI to PyPI packages

peterhessey:
Previously both the `hi-ml` and `fastmri` packages were being created as submodules, which makes very little sense in a production context.

This PR removes these submodules and instead creates an environment where both of these are instead imported as PyPI packages. In order to get the latest version of [fastmri](https://pypi.org/project/fastmri/) (0.2.0) python version >3.8 is required, and as such this PR also includes an upgrade to the python 3 version.


Tests/ML/configs/fastmri_random.py
```diff
@@ -8,18 +8,116 @@
...
 from fastmri.data.subsample import create_mask_for_mask_type
 from fastmri.data.transforms import VarNetDataTransform
 from fastmri.pl_modules import FastMriDataModule
+from pytorch_lightning import LightningDataModule, LightningModule
+
+from InnerEye.ML.configs.other.fastmri_varnet import VarNetWithImageLogging
+from InnerEye.ML.lightning_container import LightningContainer
+
+
+def create_temp_data(path: Path) -> Dict[str, Any]:
```
fepegar, line 26: We should probably add a comment with a link to the original code.

ant0nsc (approved):

PR #757, opened 2022-07-04
ENH: Upgrading package versions for security patches

peterhessey:
Upgrading various packages to patch identified critical CVEs. 

This involves updating pytorch-lightning to 1.6.4. This introduces a number of breaking changes for various components and tests which have been patched in this PR.

However, this includes two potentially problematic changes:

* `model_training.py` now no longer changes the current working directory before calling `trainer.fit()`. This was as it was causing the CIFAR SSL container tests to fail as CIFAR is downloaded to `/None/` in the head of the dir. I'm personally of the opinion that we shouldn't be changing the working directory anyway as it's a perfect recipe for annoying problems like this one (took me a long time to figure out what was going on). I think it should be the responsibility of the user running non-InnerEye models to ensure that their files are being saved to the appropriate folder, not IE-DL's job to try and guess where they're going to go.
* Updated `expected_metrics` in `test_innereye_ssl_container_cifar10_resnet_simclr()`. These regression tests were failing after updating the package versions so I updated the expected metrics to the new ones. The validation loss values had improved, the training ones worsened, neither significantly. This could affect other models but I'm not sure which models need to be checked!

Also includes an upgrade to hi-ml v0.2.5, closing #801


InnerEye/Common/common_util.py
```diff
@@ -421,3 +428,75 @@ def change_working_directory(path_or_str: PathOrString) -> Generator:
     os.chdir(new_path)
     yield
     os.chdir(old_path)
+
+
+def merge_conda_files(
```
ant0nsc, line 433: I have removed conda merging from `hi-ml` because it turned into a source of trouble. We should discuss if we can do the same for InnerEye.

InnerEye/ML/SSL/lightning_modules/ssl_online_evaluator.py
```diff
@@ -99,14 +99,14 @@ def on_pretrain_routine_start(self, trainer: pl.Trainer, pl_module: pl.Lightning
         self.evaluator.to(pl_module.device)
         if hasattr(trainer, "accelerator_connector"):
             # This works with Lightning 1.3.8
-            accelerator = trainer.accelerator_connector
+            accelerator = trainer._accelerator_connector
```
ant0nsc, line 102: please undo, this should be handled by the coade in lines 103-105.

InnerEye/ML/configs/other/fastmri_varnet.py
```diff
@@ -32,8 +32,8 @@ class VarNetWithImageLogging(VarNetModule):
     """
 
     def log_image(self, name: str, image: torch.Tensor) -> None:
-        experiments = self.logger.experiment if isinstance(self.logger.experiment, list) \
-            else [self.logger.experiment]
+        experiments = self.loggers[0].experiment if isinstance(self.loggers[0].experiment, list) \
```
ant0nsc, line 35: I don't get this change. Does the `logger` field no longer exist in new PL?

InnerEye/ML/lightning_base.py
```diff
@@ -289,7 +289,7 @@ def on_train_end(self) -> None:
         This hook is called at the very end of training. Use that to write the very last set of training and
         validation metrics from the StoringLogger to disk.
         """
-        self.read_epoch_results_from_logger_and_store(epoch=self.current_epoch)
+        self.read_epoch_results_from_logger_and_store(epoch=self.current_epoch-1)
```
ant0nsc, line 292: why is this necessary?

InnerEye/ML/model_training.py
```diff
@@ -264,11 +264,10 @@ def model_train(checkpoint_path: Optional[Path],
...
     logging.info("Starting training")
-    # When training models that are not built-in InnerEye models, we have no guarantee that they write
-    # files to the right folder. Best guess is to change the current working directory to where files should go.
-    with change_working_directory(container.outputs_folder):
-        trainer.fit(lightning_model, datamodule=data_module)
-        trainer.logger.close()  # type: ignore
+
+    trainer.fit(lightning_model, datamodule=data_module)
+    trainer.logger.close()  # type: ignore
+
```
ant0nsc, line 270: There's a particular reason for this change of workin dir, we can discuss.

Tests/ML/utils/test_model_util.py
```diff
@@ -43,10 +43,6 @@ def create_model_and_store_checkpoint(config: ModelConfigBase, checkpoint_path:
     if machine_has_gpu:
         model = model.cuda()  # type: ignore
     trainer.model = model
-    # Before saving, the values for epoch and step are incremented. Save them here in such a way that we can assert
```
ant0nsc, line 46: why deleted?

ant0nsc (changes requested):
We can probably simplify the PR by removing conda merge. Also, I have a couple of questions about your lightning changes.

InnerEye/ML/configs/other/fastmri_varnet.py
peterhessey: Yep precisely!

Tests/ML/utils/test_model_util.py
peterhessey: This is no longer necessary for the same reason as the other change of epoch value: the epochs are now incremented at a different time and these lines now break a number of tests.

InnerEye/ML/lightning_base.py
peterhessey: Epochs are now incremented at a different point, without this change many tests are broken

ant0nsc (approved):

InnerEye/Common/common_util.py
peterhessey: Agreed, will open an issue to remove this in the future but for now I think it's worth keeping in this PR to avoid making too many changes at once

InnerEye/ML/configs/other/fastmri_varnet.py
kenza-bouzid: Strange 🤔 You should be able to use logger property that return the first logger in self.loggers
image
This is defined in LightningModule which I believe `VarNetWithImageLogging` inherits from

Tests/SSL/byol/test_byol_moving_average.py
```diff
@@ -35,8 +35,8 @@ def __getitem__(self, item: Any) -> Any:
 
     byol_weight_update = ByolMovingAverageWeightUpdate(initial_tau=0.99)
     trainer = Trainer(max_epochs=5)
-    trainer.train_dataloader = dummy_rsna_train_dataloader
-    n_steps_per_epoch = len(trainer.train_dataloader)
+    trainer.train_dataloader = dummy_rsna_train_dataloader  # type: ignore
```
kenza-bouzid, line 38: why do we need to ignore typing now?

Tests/ML/utils/test_model_util.py
```diff
@@ -43,10 +43,6 @@ def create_model_and_store_checkpoint(config: ModelConfigBase, checkpoint_path:
...
-    # easily later. We can't mock that because otherwise the mock object would be written to disk (that fails)
-    trainer.fit_loop.current_epoch = FIXED_EPOCH - 1  # type: ignore
```
kenza-bouzid, line 48: You might want to delete line 35 that is now deprecated.

Tests/ML/utils/test_model_util.py
```diff
@@ -74,9 +70,6 @@ def test_create_model_from_lightning_checkpoint(test_output_dirs: OutputFolderFo
         assert config._train_output_size is None
     # method to get all devices of a model
     loaded_model = load_from_checkpoint_and_adjust_for_inference(config, checkpoint_path)
-    # Information about epoch and global step must be present in the message that on_checkpoint_load writes
-    assert str(FIXED_EPOCH) in loaded_model.checkpoint_loading_message
```
kenza-bouzid, line 78: what about checking that default epoch is still in the loading message? it must be 0 if no training has been done?
If the intent here is to check that the loading_message comprises the epoch retrieved from the checkpoint file regardless of current_epoch becoming a protected property that can't be changed outside the fit_loop. 
This can catch potential bugs in checkpoints.

kenza-bouzid (approved):
LGTM, left minor suggestions/clarification questions

Tests/ML/utils/test_model_util.py
peterhessey: Excellent point, have added this!

PR #789, opened 2022-08-17
DOC: Add a copy of the software dev document from hi-ml

ant0nsc:
- Add a copy of the software dev process that we previously had in `hi-ml` and link it
- Clean up duplicate entries around pull requests


hamurfet (approved):

docs/source/md/software_design_overview.md
```diff
@@ -0,0 +1,110 @@
+# Software Development Process
+
+This document provides a high-level overview of the software development process that our team uses for their
+repositories (most importantly [InnerEye-DeepLearning](https://github.com/microsoft/InnerEye-DeepLearning) and
+[InnerEye-Gateway](https://github.com/microsoft/InnerEye-Gateway)).
+
+For detailed guidance for developers, please refer to the [coding guidelines](coding_guidelines.md).
```
peterhessey, line 7: This is not a valid link in this repository.

peterhessey (changes requested):
~~I think it would be better to link to the hi-ml [software design overview docs](https://github.com/microsoft/hi-ml/blob/main/docs/source/coding_guidelines.md) (probably in the README) instead of copying the file over, as if any changes are made to the guidelines in the future they will need to be copied across the repos.~~

That is unless we expect the guidelines for each repo to diverge?

docs/source/md/software_design_overview.md
peterhessey: Scrap above, just seen your e-mail to Tom. For full decoupling from hi-ml the coding guidelines will also need to be copied over!

ant0nsc: I removed the link altogether. We have other docs in InnerEye that very much overlap with the coding guidelines.

peterhessey (approved):

PR #796, opened 2022-09-01
BUG: Dont update multi-node env vars for single node training

mebristo:
Environemnt variables seem to be getting set in Azure ML for single node jobs that cause our logic to fail in `set_environment_variables_for_multi_node` therefore we should avoid calling this for jobs on 1 node


ant0nsc:
@mebristo, can we not use the updated function from hi-ml? Just publish a new package version and consume it here.

mebristo:
> 

I sent you a message about this offline - the hi-ml package hasn't been updated in a while, so updating it in this PR would also require refactoring how users can provide conda files (since merge_conda_files has been removed from hi-ml in favour of a single command line arg) ... so in the spirit of a 'quick fix' to unblock PRs I felt this was the best approach

ant0nsc (approved):

fepegar (approved):

PR #804, opened 2022-09-27
ENH: Fix logging + model registration for Amulet runs

peterhessey:
Adds functionality to properly log IE-DL outputs and register trained models when running jobs through [amulet](https://amulet-docs.azurewebsites.net/).

### DIFF
--- a/InnerEye/ML/run_ml.py
+++ b/InnerEye/ML/run_ml.py
@@ -572,10 +572,11 @@ class MLRunner:
                 logging.info(f"Registering the model on the current run {run_to_register_on.id}")
             logging.info(f"Uploading files in {final_model_folder} with prefix '{artifacts_path}'")
 
-            if not is_amulet_job():
-                final_model_upload_path = final_model_folder.relative_to(Path.cwd())
-            else:
+            if is_amulet_job():
                 final_model_upload_path = final_model_folder
+            else:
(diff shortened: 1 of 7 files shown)


kenza-bouzid (approved):
LGTM

InnerEye/ML/run_ml.py
```diff
@@ -571,8 +571,12 @@ def register_model(self,
                 run_to_register_on = RUN_CONTEXT
                 logging.info(f"Registering the model on the current run {run_to_register_on.id}")
             logging.info(f"Uploading files in {final_model_folder} with prefix '{artifacts_path}'")
-            final_model_folder_relative = final_model_folder.relative_to(Path.cwd())
-            run_to_register_on.upload_folder(name=artifacts_path, path=str(final_model_folder_relative))
+
+            if not is_amulet_job():
```
ant0nsc, line 575: I find it easier to read if the condition is not inverted - you anyway cover both cases?

ant0nsc (approved):

InnerEye/ML/run_ml.py
peterhessey: Agreed + updated!

peterhessey:
> /azp run

why did this work for you but not for me? 😢

PR #814, opened 2022-10-21
ENH: Improve `rtconvert` log visibility

### LINKED ISSUE #813: Raise `rtconvert` logs to LOGGING level
Currently logs output by `rtconvert()` in score.py are only logged at the DEBUG level. This means that during training if an error occurs during DICOM conversion the logs aren't displayed. This should be changed so that error logs from DICOM conversion are available during training. 

[AB#7822](https://innereye.visualstudio.com/60ce1777-00d6-4015-82bc-488a0c00202f/_workitems/edit/7822)

peterhessey:
Closes #813

### DIFF
--- a/score.py
+++ b/score.py
@@ -213,6 +213,7 @@ def convert_nifti_to_zipped_dicom_rt(nifti_file: Path, reference_series: Path, s
     :param config: Model config.
     :param dicom_rt_zip_file_name: Target DICOM-RT zip file name, ending in .dcm.zip.
     :param model_id: The AzureML model ID <model_name>:<ID>
+    :raises: RunTimeError if rtconvert fails.
     :return: Path to DICOM-RT file.
     """
     dicom_rt_file_path = scratch_folder / Path(dicom_rt_zip_file_name).with_suffix("")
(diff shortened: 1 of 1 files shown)


score.py
```diff
@@ -229,8 +229,9 @@ def convert_nifti_to_zipped_dicom_rt(nifti_file: Path, reference_series: Path, s
         modelId=model_id
     )
     # Log stdout, stderr from DICOM-RT conversion.
-    logging.debug("stdout: %s", stdout)
-    logging.debug("stderr: %s", stderr)
+    logging.info("DICOM-RT conversion stdout: %s", stdout)
```
ant0nsc, line 232: how noisy is that actually?

score.py
```diff
@@ -229,8 +229,9 @@ def convert_nifti_to_zipped_dicom_rt(nifti_file: Path, reference_series: Path, s
...
+    if stderr != "":
+        logging.error("Errors detected during DICOM conversion: %s", stderr)
```
ant0nsc, line 234: Should we simply fail here if there's something in stderr or a non-zero exit code? Otherwise it will raise a filenotfound exception that is completely nondescript.

ant0nsc (changes requested):

score.py
peterhessey: Not at all, rtconvert only prints a single line on a successful conversion!

score.py
peterhessey: Good point!

kenza-bouzid (approved):

ant0nsc (approved):

PR #823, opened 2022-11-07
DOC: Hotfix doc links

peterhessey:
Some of the doc links in the pre-trained model page are broken, this is quick hotfix to remedy that.


peterhessey:
Bypassing branch protections for this small change as we have UCL using this docs page right now!

